PR #3, opened 2021-12-27
content: add lessons 10 and 11

johnnygossdev:
- [ ] Add better delta visual example
- [ ] Remove `move_local_x()` in examples as it already uses `delta` behind the scenes

I've combined lessons 10 and 11 into one pull request.

I'd have liked to add another practice to both lessons which used `move_local_x()`, but I couldn't get the tests to work by comparing `Vector2`.


course/lesson-3-standing-on-shoulders-of-giants/ExampleRotateSprite.tscn
```diff
@@ -16,7 +29,10 @@ margin_left = 7.0
 margin_top = 7.0
 margin_right = 1040.0
 margin_bottom = 291.0
+<<<<<<< HEAD
 scene = ExtResource( 1 )
+=======
+>>>>>>> main
```
NathanLovato, line 35: This merge conflict text will prevent the scene from loading

johnnygossdev: Ok will fix

Xananax:
I found it hard to do something like code review; it was easier and faster to intervene directly on the content.

Please check the changes and see if you agree with them. I've mostly moved things around, as well as changed a few formulations that I deemed too complex for newcomers.

One functional change I've made is to have the delta examples display rotation and framerate, so the reader can more readily follow what is going on, but this is a quick hack and doesn't look good. If we decide to keep it, we probably have to make it look nicer.

I need to say I don't think the rotation works very well for the purposes of what we're trying to demonstrate. As exercises, it's fine, but as examples, rotation makes it really difficult to notice lack of smoothness, or to realize the distance jumped from frame to frame.

A linear movement on one axis would serve much better to understand how smooth/unsmooth movement can be with or without delta.

Reporting my proposal from the chat here:

I'd see for example a car on a straight line, duplicated 3 different times:

:red_car: ------------------------------ normal
:red_car: ------------------------------ choppy (constant distance per frame)
:red_car: ------------------------------ choppy (using delta)
computer speed --o------- 

The "normal" car constantly goes from left to right and then again.
Maybe the speed slider can be a toggle between "potatoe computer" "weak computer" and "normal computer"

Another proposition to show how delta works:
Making it so user can change delta, and show the 10 frames, as well as the car/character's position, all at once.

Something like an onion skin and regular static intervals drawn on the stage.

As the user changes delta to match the actual time delta of each frame, they can see each onion skin match up visually with the markers

johnnygossdev:
I think the change in the delta example is a great idea. A straight line example would be better absolutely!

I'll make some changes and get your input again, thanks!

NathanLovato:
I'm facing a bug: I can't complete assignments in lesson 10. Even with the provided solution.

Also, the practices talk about slow movement, but on my computer, the provided argument of 0.05 to rotate leads to a fast rotation.

---------------------

Regarding the delta example in lesson 11, it's confusing to me as it is. I think there shouldn't be a reference row but only a version with and without delta. Also, there's a lot of choppiness in the motion and it's hard to understand.

If you want to get across the idea of varying time deltas, I would recommend not doing it like this as right now, both with and without delta, it just looks like neither approach works. I think showing only the character overshooting the target is enough.

In general, I think it'll be easier for students if we try to show only one thing at a time. If you want to illustrate variations in `delta`, then I'd make a different visualization. Maybe a graph of the value over time, something like that.

NathanLovato:
Fixed conflicts and made assignments pass in lesson 10

NathanLovato:
I'm going to merge now to avoid more conflicts and open a task to improve the delta presentation.

PR #78, opened 2022-01-01
chore: remove debug print statement

allkhor:
I think this added for debug purposes or printed message must be different then "sdfdsfdsf?" ?


Xananax:
Thank you, this escaped my attention :)

PR #92, opened 2022-01-03
fix: the explanation error popup not centered with different screen resolutions

allkhor:

- For bug fixes and features:
    - [x] You tested the changes.

Bug fix.

No

## New feature or change ##

**What is the new behavior?**

### DIFF
--- a/ui/components/OutputConsole.gd
+++ b/ui/components/OutputConsole.gd
@@ -22,7 +22,8 @@ func _ready() -> void:
 	_external_error_popup.set_as_toplevel(true)
 	_error_popup.set_as_toplevel(true)
 	_error_overlay_popup.connect("hide", _error_popup, "hide")
-	
+	connect("resized", self, "_on_resized")
+
 	MessageBus.connect("print_request", self, "print_bus_message")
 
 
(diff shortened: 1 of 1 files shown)


pycbouh:
You don't need to do this on the resized signal. The popup has a fixed size, you can just update the anchors in the scene (which we forgot to do when the adjusted the size of the panel). For a more dynamic solution, setting the anchors on about to show should be more appropriate.

NathanLovato:
Yup, you could just use the full screen layout for the color rect in OutputConsole.tscn, and possibly a CenterContainer with the overlay popup as a child. We generally favor declarative options for UI (-> using nodes and node settings in the scene over gdscript).

allkhor:
Setting anchor to the scene it's first was I do, but the ColorRect is child of container and when visibility manually changed anchor was gonna reset to zero. I think on desktop platform when explanation popup was opened and user has change the window size, popup must be recalculated to be centered on the screen like other UI elements fit free space. I can update anchor on ColorRect, but this not covered case when user change the windows size, it's not affected html export, but only desktop, WDYT?

NathanLovato:
Got it, I hadn't noticed it was in the output console scene (there's another somewhere else). The PR fixes the issue so let's merge it. Thanks much for the help once more!

PR #120, opened 2022-01-06
content: add lesson 13

johnnygossdev:
- [x] Write lesson
- [x] Create lesson visuals
- [x] Create lesson quizzes
- [x] Create lesson practices
- [x] Add better visuals to practices
- [x] Improve visuals to be more explicit?

Adds lesson 13: Conditions.

See [the curriculum](https://github.com/GDQuest/learn-gdscript/blob/main/course/README.md#lesson-13-conditions) for more information.


Xananax:
Not much to say, this is good! Pleasant to read and clear.

One potential improvement would be to explain shadowing somewhat, specially that it is likely to be a mistake people can do in exercise 2.

image

Alternatively, or additionally, we can also add it to the important errors to explain.

Xananax:
I brought in some minor changes. It was mostly good.

A few comments:
1. we're missing some sort of graphic in the exercises. Maybe show the character with life increasing or diminishing (is this planned?)
2. I found that example with only `heal()` to be really confusing. I'd rather we do this:

image

This is part of a larger discussion we're having about being explicit vs implicit, this is just an additional example to add to the pile.

Side note: we probably want to add this error to errors that we should explain in priority:

image

johnnygossdev:
I've added improvements to the lesson and practices based on your feedback @Xananax. If there's nothing else, it should be good to merge.

Xananax:
Looks good to me

Xananax (approved):
Good to go

Issue #123, opened 2022-01-06
Wrong wording in "Early beta release"-infobox


Dinock:
**Issue description:**
In the last sentence of the "Early beta release"-infobox, which appears at the beginning of the course, there seems to be a typo/wrong wording.

Right now it is written:
`we'll use your reports and feedback to improve the app, both your you and everyone else.`

It probably should be:
`we'll use your reports and feedback to improve the app, both for you and everyone else.`
(One `your` has been changed to a `for`.)

Unfortunately I could not find the right file. Otherwise I would have tried to submit a PR by myself.

**Screenshot of the current infobox:**
![Bildschirmfoto vom 2022-01-04 22-38-53](https://user-images.githubusercontent.com/54438803/148427033-f0eb8895-472e-4dd5-b8e3-ac6339ad44b3.png)

duianto:
@Dinock This seems to be where that text is defined.

Two instances of: `both your you`
were found when I searched in the top left Github search box (`In this repository`):

https://github.com/GDQuest/learn-gdscript/blob/8291d60a17c0f64688367d07bbfe939dabc45f68/ui/welcome_screen/WelcomeScreen.tscn#L346

https://github.com/GDQuest/learn-gdscript/blob/8291d60a17c0f64688367d07bbfe939dabc45f68/ui/welcome_screen/WelcomeScreen.tscn#L353

PR #126, opened 2022-01-07
fix: typo on end screen

Dinock:
This small PR fixes wrong wording in the welcome screen. 
It changes the word "your" to "for" in the "Early beta release"-infobox.

This should fix issue #123.

Thanks to @duianto for helping me out! :)


NathanLovato:
Thanks! Note that in cases like this one, you can create the PR directly without bothering with issues. That'll save you a little bit of time.

PR #151, opened 2022-01-18
content: add lesson 9

johnnygossdev:
This lesson talks about addition and subtraction in the context of altering the robot's health.

See [the cirriculum](https://github.com/GDQuest/learn-gdscript/blob/main/course/README.md#lesson-9-adding-and-subtracting) for more information.


Xananax (commented):
Looks good to me, nothing to report except:

in the first exercise, it's possible to pass the checks with the code

```gdscript
func take_damage(amount):
    health -= 50
```

Which we might want to disallow, either by regexing `amount` and making sure it's used on the line, or by supplying different damage values in repeated calls to `take_damage`.

I'm leaving this as a comment rather than a "request changes" because I don't think it matters overmuch, feel free to address this or just merge (we can file this as a bug for later too)

johnnygossdev:
Captured this case, thanks!

PR #156, opened 2022-01-20
content: add lesson 15

johnnygossdev:
To Do:
 - [x] Write lesson
 - [x] Add practices

This adds lesson 15: Modulo.

It shows how to use modulo through the context of cycling traffic lights, finding even numbers, and rolling random dice.

See [the curriculum](https://github.com/GDQuest/learn-gdscript/blob/main/course/README.md#lesson-15-modulo) for more information.

image


NathanLovato:
Good job overall. I made a number of edits. Most important ones:

1. The sliders in the first modulo example now are constrained to one another. You can't make the modulo larger than the number anymore.
2. Using `return` in functions. I'll add this important concept to the curriculum, we should cover that before this lesson. We shouldn't invite the students to use properties to store the result of functions.

PR #187, opened 2022-01-27
Explain Jump

Feniks-Gaming:
Expanded lesson text to clarify what function jump() does

### DIFF
--- a/course/lesson-5-your-first-function/practice-kGx0c7DD.tres
+++ b/course/lesson-5-your-first-function/practice-kGx0c7DD.tres
@@ -10,7 +10,7 @@ goal = "You have a function to draw one square: [code]draw_square()[/code]. Use 
 
 We already created [code]draw_square()[/code] for you. Create a function named [code]draw_three_squares[/code] that calls [code]draw_square()[/code] three times.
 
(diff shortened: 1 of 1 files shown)


course/lesson-5-your-first-function/practice-kGx0c7DD.tres
```diff
@@ -10,7 +10,7 @@ goal = "You have a function to draw one square: [code]draw_square()[/code]. Use
 
 We already created [code]draw_square()[/code] for you. Create a function named [code]draw_three_squares[/code] that calls [code]draw_square()[/code] three times.
 
-If you just call the function, all three squares will overlap. To stack them diagonally, call [code]jump(300, 300)[/code] between two calls to [code]draw_square()[/code]."
+If you just call the function, all three squares will overlap. To stack them diagonally, call [code]jump(300, 300)[/code] between two calls to [code]draw_square()[/code]. Function [code]jump(300, 300)[/code] will change turtle position by 300 pixels to the right and 300 pixels down without drawing any lines. "
```
NathanLovato, line 13:
```suggestion
If you just call the function, all three squares will overlap. To stack them diagonally, call [code]jump(300, 300)[/code] between two calls to [code]draw_square()[/code]. Calling [code]jump(300, 300)[/code] makes the turtle jump by 300 pixels to the right and 300 pixels down without drawing any lines."
```

PR #196, opened 2022-01-27
fix: decapitalized after semicolon

MaaaxiKing:


NathanLovato:
Capital letter after a semicolon is the US syntax. We write in US English.

PR #218, opened 2022-01-31
fix: throttle scroll events

### LINKED ISSUE #114: Scrolling is too fast on some touchpads
## **Describe the bug**
The scrolling on the Synaptic's touchpad is too fast. This is not critical but very annoying. You have to use a small scroll bar,
which is VERY inconvenient on a laptop without a mouse.

## **To Reproduce**
Just scroll on the touchpad 

**Browser:** Firefox 95.0.1
**OS:** Manjaro with Gnome Desktop

NathanLovato:
This should help fix issues with touchpads firing too many scroll events on some web browsers.

Should fix #114 but needs testing.

### DIFF
--- a/addons/SmoothScroll/SmoothScrollContainer.gd
+++ b/addons/SmoothScroll/SmoothScrollContainer.gd
@@ -19,6 +19,7 @@ const ACCELERATE_DISTANCE_THRESHOLD := 120.0
 # Used to multiply the scroll speed as the target scroll gets farther away than
 # ACCELERATE_DISTANCE_THRESHOLD.
 const SPEED_DISTANCE_DIVISOR := 200.0
+const TIME_MSEC_BETWEEN_SCROLL_EVENTS := 15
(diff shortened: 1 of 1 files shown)


NathanLovato:
@Xananax could you test this one and tell me if it helps on Chrome for you? You can control the throttling in SmoothScrollContainer.gd by changing `TIME_MSEC_BETWEEN_SCROLL_EVENTS`.

Xananax:
On chrome, it works perfectly well. On Firefox, I get occasional jumps of about 20px up or down from time to time. It's mainly noticeable while testing, or if someone wants to scroll to bottom very fast. If someone's just reading and scrolling at reading speed, it will seem a bit glitchy sometimes, but it's fine.

I don't think it's worth exploring this glitch, there are too many confounding factors (hardware, performance, javascript event handling, Godot, ...). I deem the current behavior satisfying.

PR #231, opened 2022-02-01
fix: typo "minute" with "moment"

duianto:

Fixed a typo.

Before:
>At the minute, the rotation of the robot is frame-dependent.

Changed: `minute` to `moment`

After:
>At the moment, the rotation of the robot is frame-dependent.


NathanLovato:
Thanks

Issue #15, opened 2021-12-28
Animate quizzes


NathanLovato:
When giving the right answer, the quizz panel pops instantly. It’d be nice to animate the height. We could also animate the panel when showing a hint, see Consider adding optional hint to quizzes

mathmods:
I'm working on this enhancement.

I'm considering using the existing tween object on a ColorRect (invisible by modulation) to change the size of the container without jumps. (Also adding a modulation tween to the resultview to ensure text doesn't "pop").

> We could also animate the panel when showing a hint, see Consider adding optional hint to quizzes

With this I'll consider making a "_switch_view" function that takes two views and animates the transition between the elements.

NathanLovato:
Thanks for the help! Tweening the rect_min_size of some element would definitely be the way to go. Youĺl likely have to tween the size of the root container of the quiz UI so it properly moves elements around it in the lesson UI.

You can also set some container node to Clip Content during the transition to animate the container size. I think Revealer.tscn is a good example of how to approach UI animation like that.

And I see the referenced issue in my top post didn´t get linked correctly, itś #16

mathmods:
Thanks for the tips! I'll look into that.
Clipping content will probably look better in code (and practice) than the thing I attempted last night (yielding to timers to hide and show the items).

And I now see how the help is supposed to work, I'll deal with that once I've gotten it to actually change size properly.
Might add an internal function that just checks the size of it's own contents and then animates to fit. (Upon size change of the margin container perhaps)

PR #234, opened 2022-02-01
improvement: Quiz resize animations

mathmods:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #15 

Animates resizes of Quiz content blocks. Reacts upon any change of choice view or result view and will proceed to fit (if they are visible).

No

## New feature or change ##

Instantaneous jumps to new size.

**What is the new behavior?**
Animated size changes of rectangle size to fit the current view.

**Other information**
Letters sometimes gets a black line over them for no apparent reason. Not sure if this is created on my own pc running this or if this is introduced by BoundaryControl. Affects text on quiz answer elements and quiz label.

You can see the quiz unfolding itself if you scroll fast enough. (I couldn't find a way to set the correct size without waiting for the views to update with the loaded content.)

Some answers with long text might wrap around once selected and increase the size of the view creating a slightly weird looking reaction by the container. (Not breaking anything, looks normal after .5 seconds)

Will work with hints when implemented.

### DIFF
--- a/ui/screens/lesson/UIBaseQuiz.gd
+++ b/ui/screens/lesson/UIBaseQuiz.gd
@@ -12,33 +12,42 @@ const OUTLINE_FLASH_DURATION := 0.8
 const OUTLINE_FLASH_DELAY := 0.75
 const ERROR_SHAKE_TIME := 0.5
 const ERROR_SHAKE_SIZE := 20
+const FADE_OUT_TIME := 0.15
+const FADE_IN_TIME := 0.15
+const INTERPOLATION_TIME := 0.3
 
 export var test_quiz: Resource
 
 var completed_before := false setget set_completed_before
 
 onready var _outline := $Outline as PanelContainer
-onready var _question := $MarginContainer/ChoiceView/QuizHeader/Question as RichTextLabel
-onready var _explanation := $MarginContainer/ResultView/Explanation as RichTextLabel
-onready var _content := $MarginContainer/ChoiceView/Content as RichTextLabel
+onready var _question := $MarginContainer/BoundaryControl/ChoiceView/QuizHeader/Question as RichTextLabel
+onready var _explanation := $MarginContainer/BoundaryControl/ResultView/Explanation as RichTextLabel
+onready var _content := $MarginContainer/BoundaryControl/ChoiceView/Content as RichTextLabel
 onready var _completed_before_icon := (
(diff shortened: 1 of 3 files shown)


ui/screens/lesson/UIBaseQuiz.gd
```diff
@@ -8,37 +8,47 @@ const ERROR_OUTLINE := preload("res://ui/theme/quiz_outline_error.tres")
...
+onready var _error_tween := $ErrorTween as Tween
+onready var _size_tween := $SizeTween as Tween
+onready var _help_message := $MarginContainer/BoundaryControl/ChoiceView/HelpMessage as Label
 
 var _quiz: Quiz
 var _shake_pos: float = 0
+# For animating changing the size of the container
+var _previous_rect_size := rect_size
+var _next_rect_size := Vector2.ZERO
+var _percent_transformed: float = 0
```
NathanLovato, line 51: You can use := 0.0 to get float type inference

mathmods:
Hmm. It seems to only be present if the content is set to something. I got it to print the sizes of everything in the tree and it practically told me that nothing was using that space. Essentially, choice_view isn't updating it's own size to match? Or rather it just keeps it size? I'll mess about with this to see if it leads me somewhere

mathmods:
Ok. Got the gap to become smaller by forcing a resize, but it's still there. Approximately 1 line instead of 2~3.
Removing the expand flag from vertical makes the gap appear at the bottom and not in the middle which is confusing me to no end.

I'm considering just cheating by removing the "expand" flag on the rich label and then assume a smaller size if content is visible.

mathmods:
Ok, I noticed this is affecting different lessons in different ways. With the newest commit: any content with 1 line will display an additional line just beneath it. Any content larger than that and it's normal. 

In other words; I see 3 possible solutions:

* 1 line once in a while isn't that bad, just commit this as it is.
* Try and add an empty line before the content so that it takes some distance from the header, instead of the answers having distance to the content.
* Try and detect 1 line contents and "cheat" as I mentioned in the previous comment.

mathmods:
Ok, I have found a better solution that completely removes it.
Sadly it will change how the animation works since it moves the boundary to become the parent of the margin container and not vice versa.

It will take some time, but I'm going to make it work properly.

NathanLovato:
Thanks for taking the time, we truly appreciate it. I'll be there to review and merge your new PR when it's ready.

PR #240, opened 2022-02-01
fix: missing space between a nr and modulo sign

duianto:

Fix typos.

Before: `So 11% 4 is 3`
After: `So 11 % 4 is 3`

And:
Before: `For example, 1% 2 gives you 1.`
After: `For example, 1 % 2 gives you 1.`


NathanLovato:
Thanks much for all the fixed typos! Please note you don't need to fill the commit description in those cases as the change is self explanatory. We only use commit descriptions for complex features, for example, to explain why or how we ended up with this or that algorithm or implementation.

PR #247, opened 2022-02-02
improvement: animate quiz resizing

mathmods:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #15 

Animates resizes of Quiz content blocks. Reacts upon any change of choice view or result view and will proceed to fit (if they are visible).
Almost the same as #234 except with proper spacing on some content.

This PR changes the Quiz Base UI to use a Control node with clip content on to contain margin containers. It reacts to the minimum size changes of the margin containers to determine the size it's supposed to animate to.

No

## New feature or change ##

Instantaneous jumps to new size.

**What is the new behavior?**
Animated size changes of rectangle size to fit the current view.

**Other information**
There are now 2 margin containers to ensure that one view doesn't affect the size of the other view. This helps remove guesswork from resizing animations. (And massively improves readability of the code.)

### DIFF
--- a/ui/screens/lesson/UIBaseQuiz.gd
+++ b/ui/screens/lesson/UIBaseQuiz.gd
@@ -47,6 +47,7 @@ var _shake_pos: float = 0
 var _previous_rect_size := rect_size
 var _next_rect_size := Vector2.ZERO
 var _percent_transformed := 0.0
+var _animating_hint := false
 
 
 func _ready() -> void:
@@ -56,6 +57,7 @@ func _ready() -> void:
 	_skip_button.connect("pressed", self, "_show_answer", [false])
 	connect("item_rect_changed", self, "_on_item_rect_changed")
 	
+	_help_message.connect("visibility_changed", self, "_on_help_message_visibility_changed")
 	_choice_container.connect("minimum_size_changed", self, "_on_choice_container_minimum_size_changed")
 	_result_container.connect("minimum_size_changed", self, "_on_result_container_minimum_size_changed")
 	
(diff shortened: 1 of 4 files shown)


mathmods:
@NathanLovato While this new attempt does address the issue with the last one, I'm wondering what looked best:

* The new one where the clip border is at the bottom of the UI, essentially making no boundary between the edge and other content during resizing.
* The old one where the clip border was at the margins of the container, making text appear/disappear inside the UI avoiding temporary no-margin times.

Personally I found the old one better due to the buttons on the bottom looking quite bad when pushed to the edge of the UI.
I have already implemented it (on local repository), but I'm getting the feeling that it should probably be done as a new type of container than having everything inside 2 controls whose sole purpose is to clip content.

NathanLovato:
I noticed the submit button overflowing when some answer gets selected and takes more space. I think in those cases there should be no animation, because text wrapping is instant, causing an instant change in the parent container(s).

So to me the way it works now is good when submitting or skipping the quiz, but it shouldn't animate like that when selecting answers.

mathmods:
Hmm. I did notice that as well. I can forcefully change the size if the detected change comes from the answer list and not anything else. (Or just disable animations for the choice container - though showing a hint would become instant as well in that case.)

mathmods:
@NathanLovato The latest commit should address this. It looks for if the help labels visibility changed, if not just resize instantaneously.

NathanLovato:
Thank you so much for your contribution! It's a most welcome improvement.

Issue #220, opened 2022-01-31
Lesson 9: health progress bar shrinks on first `Reset`


duianto:
**Is your feature request related to a problem? Please describe.**

In the first and second code examples in lesson 9:
https://gdquest.github.io/learn-gdscript/#course/lesson-9-adding-and-subtracting/lesson.tres

Pressing the `Reset` button shrinks the health progress bar slightly.
Compared to when the lesson page was (re)loaded.

First example:
Before: The progress bar ends to the right of the 5 in `health = 75`:

![msedge_wSg4npiQ2e](https://user-images.githubusercontent.com/13420573/151856105-88c5e033-0d30-49b8-b092-71a2a4e10fd1.png)

After: The progress bar ends under the 5 in `health = 75`:

![msedge_CeLrTzF2LD](https://user-images.githubusercontent.com/13420573/151856310-0a4aa8e5-6493-4666-a07d-0b13d9251a59.png)

The same thing happens in the second example:

Before: The progress bar ends under the first letter `h` in `health = 25`:

![msedge_2FVJcEs71S](https://user-images.githubusercontent.com/13420573/151856543-e7113023-f4cb-4404-b9ba-390aa5a1902e.png)

After: The progress bar ends below the left side of the first letter `h` in `health = 25`:

![msedge_Ra0asO6iRi](https://user-images.githubusercontent.com/13420573/151856639-dba61bf3-586f-44e4-8be1-ca472255189e.png)

**Describe the solution you'd like**
The progress bar's initial size should probably match its reset size.

**Note**

This is most likely a low-priority enhancement.
It doesn't affect the lesson's functionality.
It's just an observation that pressing the `Reset` button,
resizes the progress bar once, from its initial size.

mathmods:
While indeed the original rect size value is wrong (it's calculated differently), it also calls the same function that reset calls to update it again.

It looks to me that something is blocking the tween from animating to the correct first size?

An easy fix would be to just update the original value to the correct one by using the same formula (190.0 * health / max_health). - And I would prefer using the empty health bars size as reference instead of the magic constant 190.0.

PR #248, opened 2022-02-03
fix: lesson 9 and 13 health bars now show proper values on ready

mathmods:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #220 

Ensures that the health bars in lesson 9 and 13 show correct values when loaded in. 
For some reason the tween doesn't update it on load so it has to be set manually at first.

No


NathanLovato:
Thanks for the contribution!

Issue #258, opened 2022-02-04
Unclear error message when using quoted number in lesson 8


AimVaguely:
**Describe the bug**
In [lesson 8 practice](https://gdquest.github.io/learn-gdscript/#course/lesson-8-defining-variables/practice-lkGx0c7D.tres) if you enter `var health = "100"` the error message in unclear

**To Reproduce**
Steps to reproduce the bug:
1. Go to [lesson 8 practice](https://gdquest.github.io/learn-gdscript/#course/lesson-8-defining-variables/practice-lkGx0c7D.tres)
2. Enter `var health = "100"` into text box
3. Press run

**Observed behaviour**
The error it gives is `ERROR Health variable's values is 100; It should be 100`. This error seemingly makes no sense, and I imagine a though process along the line of "You are saying it should be 100, but it is 100, so why are you telling me it is wrong then?" might happen.

**Expected behaviour**
It should either:
- give error `ERROR Health variable's values is "100"; It should be 100` or something similar to clearly state that it is a quote, not a number
- Inform user that values in quotation marks is not the same thing as a value without though some other means (see more info in additional context)

**Information about your device (please complete the following information):**
 - Operating System: Linux x64 (Manjaro)
 - Browser: Firefox 96.0.3 (64-bit)

**Additional context**
This might not be a real problem for someone learning to program, they might just not think to put quotation marks around it (you did show doing this previously, so it might be possible) I did this because I was curious how the error looked if I did so, and it looks like it might cause confusion, so I reported it. If no testers that are in the intended target audience (complete beginners to programming) then this might not be worth solving. I have a feeling this is something to do with how Godot output the message, and if it is, then there might need to be a special case with an extra error message to explain why the error message seemingly makes no sense.

NathanLovato:
I don't think that's a problem. It seems you're not a beginner and you were pushing the app's error checking.

What we need here are actual things people in the target audience do. At this stage, they haven't even learned about strings, so that shouldn't be a problem.

If beginners start to do things like those, then we should probably add some general type-checking mechanism. For this one, I'll try to leverage type hints perhaps, that should be enough.

AimVaguely:
[...]

PR #261, opened 2022-02-04
fix: lesson 8 practice 1 error message will now show that the variable is a string if it's wrong

mathmods:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #258 

Adds quotation marks '"' at the beginning and end of any string element should it be detected as a variable in the practice of lesson 8.

No


NathanLovato:
I think it's not the right approach for this one. See this comment for more information: https://github.com/GDQuest/learn-gdscript/issues/258#issuecomment-1030576064

In short, I think we should use type hints and have the parser raise a type error.

Nonetheless, thanks much for the contribution!

Issue #256, opened 2022-02-04
Drawing Multiple Squares lesson incorrect error


AimVaguely:
**Describe the bug**
If you call `draw_square()` more than once in a row without `jump(300,300)` between them, it tells you the square is not of length 200 pixels, instead of telling you that you drew too many, or that the square has too many corners.

**To Reproduce**
Steps to reproduce the bug:
1. Go to [Practice - Drawing Multiple Squares](https://gdquest.github.io/learn-gdscript/#course/lesson-5-your-first-function/practice-kGx0c7DD.tres)
2. Click on Solution
3. add an extra `draw_square()` after line 17 with the same indenting as line 17
4. Click run
5. Get `ERROR Shape number 3 is not a square of length 200 pixels. Did you change the draw_square() function`

**Expected behavior**
It reports that you drew too many squares, or that the square has too many corners.

**Information about your device (please complete the following information):**
 - Operating System: Linux x64 (manjaro)
 - Browser: Firefox 96.0.3 (64-bit)

PR #262, opened 2022-02-04
fix: added further checks to lesson 5 practice 2 with better error messages

mathmods:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #256 

Adds a check to practice 2 in lesson 5 that will determine if there are too few or too many points in the polygon, showing a similar, yet more precise error message. Note that the user must have made 3 polygons for this to happen.
Most of these boil down to the draw_square function having been rewritten or draw square being called more than once in a row (with 3 successful polygons). Though I doubt the second reason would happen normally, hence why the error message asks for if the draw square function has been rewritten, and not if it was called twice.

No


NathanLovato:
Thanks much!

PR #274, opened 2022-02-05
improvement: added warning to student when attempting to leave a Practice screen with edited code.

mathmods:

- For bug fixes and features:
    - [ ] You tested the changes.

Related issue (if applicable): #9 

If the student attempts to leave the practice (through any means given to them), it will ask for confirmation before performing the navigation. This happens if the code in the practice has changed and is not finished.

Currently maybe. 
Not properly tested yet (and I need to search the code for signal listeners - might want to add a few extra listeners to Events if I find anything).

## New feature or change ##

The practice screen immediately proceeds forward with the navigation request even if this means the student loses data.

**What is the new behavior?**
The practice screen will ask for confirmation (with warnings) before proceeding.

**Other information**
Currently, it is set to warn students that they will lose their data if they change practice, but they won't. The data is still loaded into history, so if the back button is pressed they will return to the last practice with the data intact.

It will also warn students who finished the practice, but stayed and changed the code later.

It feels like this commit adds more load time to every practice, though I'm not exactly sure *how* (Since nothing about the actual loading has been changed).

### DIFF
--- a/autoload/Events.gd
+++ b/autoload/Events.gd
@@ -13,6 +13,7 @@ signal practice_completed(practice)
 signal practice_next_requested(practice)
 signal practice_previous_requested(practice)
 signal practice_requested(practice)
+signal breadcrumbs_navigation_requested(path)
 signal course_completed(course)
 
 signal settings_requested
--- a/ui/UINavigator.gd
+++ b/ui/UINavigator.gd
@@ -4,9 +4,12 @@ extends PanelContainer
 signal transition_completed
 signal return_to_welcome_screen_requested
 
+enum NAVIGATION_REQUEST_TYPE {BACK, OUTLINER, HOME, BREADCRUMBS, NEXT_PRACTICE, PREVIOUS_PRACTICE, TO_PRACTICE}
+
 const CourseOutliner := preload("./screens/course_outliner/CourseOutliner.gd")
 const BreadCrumbs := preload("./components/BreadCrumbs.gd")
 const LessonDonePopup := preload("./components/popups/LessonDonePopup.gd")
+const LeaveUnsavedPracticePopup := preload("./components/popups/LeaveUnsavedPracticePopup.gd")
 
 const SCREEN_TRANSITION_DURATION := 0.75
 const OUTLINER_TRANSITION_DURATION := 0.5
@@ -35,6 +38,7 @@ onready var _report_button := $Layout/Header/MarginContainer/HeaderContent/Repor
 onready var _screen_container := $Layout/Content/ScreenContainer as Container
 onready var _course_outliner := $Layout/Content/CourseOutliner as CourseOutliner
 
+onready var _leave_unsaved_practice_popup := $LeaveUnsavedPracticePopup as LeaveUnsavedPracticePopup
 onready var _lesson_done_popup := $LessonDonePopup as LessonDonePopup
 
 onready var _tween := $Tween as Tween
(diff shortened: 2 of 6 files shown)


pycbouh:
> Currently, it is set to warn students that they will lose their data if they change practice, but they won't. The data is still loaded into history, so if the back button is pressed they will return to the last practice with the data intact.

Yes, we only need to have a warning when it actually will happen: back navigation, or going to outliner. Home button is not available in lessons, and forward navigation doesn't cause data loss.

You also need to account for in-browser navigation. The reason why requests go directly to the NavigationManager is because this is a unified interface for in-app and in-browser navigation requests. As such, I think we may need to devise a NavigationManager pre-navigation check that would detect if the requested action would result in the current screen being unloaded and ask that screen if it wants to prevent it.

mathmods:
I was worried about that, and originally I thought it wasn't feasible to do it in NavigationManager due to NM not knowing if UINavigator was even loaded in (, if anyone was listening to it's signals). After giving it some more thought I realize it's probably quite easy - In NavigationManager: Upon any change in navigation, check if the current site is a Practice (effectively confirming UINavigator's presence). If it is a Practice and it is a possibly lossy navigation (outliner, back or breadcrumbs - breadcrumbs is a little difficult...) signal lossy_(back/outliner/direct)_navigation_requested to UINavigator and await being called back for confirm (probably just a bool is_confirmed in the navigation function call and being recalled by UINavigator with it set to true).

I'll see if I can implement it tomorrow.

mathmods:
I just realized breadcrumbs also isn't a lossy site change, it literally loads an entirely new page onto the screen stack, but it also isn't exactly intuitive that you need to use the back buttons to return to the practice (as using any of the practice buttons will load an entirely new practice, not the currently active one).

pycbouh:
> I just realized breadcrumbs also isn't a lossy site change, it literally loads an entirely new page onto the screen stack, but it also isn't exactly intuitive that you need to use the back buttons to return to the practice (as using any of the practice buttons will load an entirely new practice, not the currently active one).

That is done on purpose. Same as opening a new page in the browser, breadcrumbs or not, will open _a new page_ that is similar to an old page, but is new.

pycbouh:
> I was worried about that, and originally I thought it wasn't feasible to do it in NavigationManager due to NM not knowing if UINavigator was even loaded in (, if anyone was listening to it's signals). After giving it some more thought I realize it's probably quite easy - In NavigationManager: Upon any change in navigation, check if the current site is a Practice (effectively confirming UINavigator's presence). If it is a Practice and it is a possibly lossy navigation (outliner, back or breadcrumbs - breadcrumbs is a little difficult...) signal lossy_(back/outliner/direct)_navigation_requested to UINavigator and await being called back for confirm (probably just a bool is_confirmed in the navigation function call and being recalled by UINavigator with it set to true).

I don't think you need `UINavigator` at all, other than to act as an intermediate perhaps between the loaded page and the manager. The page should be the one to warn about a potentially lossy navigation, and the page should be the one to bring the popup. It's all based on the context of the practice screen, so the practice screen should decide. `NavigationManager` should only ask the current screen "Hey, I'm going to unload you, is there anything you want to do?" and the `UIPractice` screen would check if the editor is dirty and if it is, add a warning and confirmation, then reconfirm the request to the NM, or deny it.

mathmods:
Thanks for the feedback! 

The step-by-step list helps a lot. Though UINavigator will have to mediate either by listening and pushing signals, or telling the screens whether or not they are the current screen.

I'm currently leaning towards the solution of asking the screens to listen themselves. That way it is probably easier to deal with having to check multiple screens in the future (let the screens deal with it, UINavigator doesn't care too much).

PR #276, opened 2022-02-06
improvement: have a single Robot scene for visuals and practices

### LINKED ISSUE #249: Have one robot scene we can use throughout the course
**Issue description:**
We use the robot a lot throughout the course.

I'd like to refactor visual scenes across multiple lessons to use a common robot scene. It would have all the animations we've made so far (some are place holder right now, I can improve them too).

The robot would have the following animations:
- wave
- level up
- take_damage
- heal
- jump

This'll also improve visuals in lessons. For example, when the students makes the robot visible in lesson 3, we can play the wave animation too, rather than having static sprites.

Should we need any more animations, we can easily add to the robot scene.

### LINKED ISSUE #260: ExampleRotateSprite.tscn in Lesson 3 doesn't show
The scene doesn't open and is not shown in the lesson because the file is corrupted.

image

image

johnnygossdev:
This PR replaces each visual that uses the robot sprites with a dedicated robot scene which is easier to maintain and add to. 

fixes #249
fixes #260

Previously, there were several different versions of the robot because many of the lessons were created in parallel.

There are currently 6 animations, some placeholder:
- idle (default)
- jump
- heal
- damage
- level
- saying_hi

The robot uses an AnimationTree to control its animations.

image

To play an animation, use the following. The robot will automatically go back to its "idle" animation.
```
onready var _animation_tree := find_node("AnimationTree")

func _ready():
	_animation_tree.travel("jump")
```

The AnimationTree has a helper script to allow the above which makes moving to each animation easier.
```
## RobotAnimationTree.gd
extends AnimationTree

signal animation_finished

onready var _state_machine = self["parameters/playback"]
onready var _animation_player = get_node(anim_player) as AnimationPlayer

func _ready() -> void:
	_animation_player.connect("animation_finished", self, "_on_animation_finished")

func travel(animation_name: String) -> void:
	_state_machine.travel(animation_name)

func _on_animation_finished() -> void:
	emit_signal("animation_finished")
```

In Godot, currently AnimationPlayers stop emitting signals when controlled by an AnimationTree, so each animation emits the _animation_finished_ signal via the method track which feeds up to the AnimationTree.

image

This allows us to wait on animations before continuing in practices.
```
## IncreaseMaxHealth.gd
# ...
func _run():
	for i in range(2):
		level_up()
		_animation_tree.travel("level")
		yield(_animation_tree, "animation_finished")
	Events.emit_signal("practice_run_completed")
```

I tested all of the visuals and practices, so everything _should_ just seemlessly work.


NathanLovato:
I quickly went through the scenes and everything seems to be working well. I'll have a tester go through all lessons again soon too. Thanks for your work!

Issue #272, opened 2022-02-05
Change the full-screen button's icon when toggling full-screen from the browser (and not from the app)


NathanLovato:
When the user toggles full-screen mode in their browser from outside the app's context, Godot doesn't know it and we can't update the full-screen toggle button.

![image](https://user-images.githubusercontent.com/12694995/152658339-a1ad4c2a-ee38-4ed2-98d3-b60e3cc55776.png)

This makes it stay in the same state until the user clicks it or toggles full-screen from the app.

The behavior may depend on the browser, but for example in brave, if a page is full-screen and you press F11, the browser itself will catch the event and toggle full-screen - it'll bypass the app.

@Xananax do you know if there's a way to get a full-screen toggled event from the browser in this case? In commit d92a7848d115a6d32cd863ad541d3c71a892c92d I worked around OS.window_fullscreen, which was giving unreliable results in the browser and caused bug #226. We could hook into a JavaScript event, if it exists, to emit the Events.fullscreen_toggled signal when toggled from outside the app.

pycbouh:
https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenchange_event

Xananax:
I'll do the wiring

NathanLovato:
Thanks!

Xananax:
The wiring is available in #278: Browser does tell Godot fullscreen happened/was exited.

However, in view of #208, additional changes are proposed in the same PR. If the additional changes are to be rejected, then we only need to pull out the javascript files (`bootstrap.js` and `bootstrap.d.ts`), and add this code to `Globals.gd`:

```gdscript
var _js_on_fullscreen_ref: JavaScriptObject

func _init() -> void:
	if not OS.has_feature('JavaScript'):
		return
	_js_on_fullscreen_ref = JavaScript.create_callback(self, "_on_js_fullscreen")
	var GDQUEST := JavaScript.get_interface("GDQUEST")
	# warning-ignore:unsafe_property_access
	GDQUEST.events.onFullScreen.connect(_js_on_fullscreen_ref)


func _on_js_fullscreen(_args: Array) -> void:
	var is_it: bool = _args[0];
	print("fullscreen set from JS: %s"%[is_it])
	Events.emit_signal("fullscreen_toggled")
```

Issue #226, opened 2022-02-01
max_screen icon issue


survivant:
if I check at this screen (when the course is complete)
![image](https://user-images.githubusercontent.com/191879/151899725-81618718-6370-4e0a-be9e-7b48ebaa21d5.png)

if I click on the icon, the app will take the full screen
![image](https://user-images.githubusercontent.com/191879/151899775-e95e7df0-deb9-4b61-9e6b-2572f0c795f3.png)

but if I click again on the icon, the screen return to normal, but the icon changed
![image](https://user-images.githubusercontent.com/191879/151899812-e3fdecd1-1cef-420b-b323-8e079d2407a1.png)

NathanLovato:
This only happens in the browser. I checked the code, it's supposedly correct and behaving as intended on the desktop. But once it gets to the browser, it fails, suggesting that the value of `OS.window_fullscreen` doesn't work the same as on desktop. For now, I don't know how to fix this.

survivant:
is it possible to detect if you are in a browser, if that's the case, you could save the screen size value before clicking on the button and use those values when you click again.

NathanLovato:
I found a workaround. We still need to get some kind of event or something from the browser when the user toggles fullscreen from outside the app - I'm unsure we can, but let's see, I'll open a separate task

Issue #208, opened 2022-01-29
Fullscreen in Chrome/Firefox disturbs the button detection


ShalokShalom:
**Describe the bug**
When I switch into fullscreen on Firefox 96 on Linux X11, buttons become unresponsive when I hover over them with the cursor and instead they become clickable, when they shouldn't. 

The whole detection of where buttons are is messed up. 

**Expected behavior**
Buttons are clickable when I hover over them in fullscreen. 

**Screenshots**
![Screenshot_20220129_115023](https://user-images.githubusercontent.com/6344099/151658289-2242db63-40d5-4e7e-b829-201099b5a08d.png)

**Information about your device (please complete the following information):**
 - Operating System: KaOS Linux
 - Browser Firefox

**Additional context**
The text and the frame of the button in the applied screenshot are supposed to light up white.

ShalokShalom:
It works just fine in non-fullscreen mode.

NathanLovato:
We have a couple of browser-specific bugs like this one. I don't know why they occur and if we can do anything about them - it may be an issue on Godot's side. Although it could also have to do with our HTML/CSS code. We'll have to investigate.

MattHorn25:
Just adding to the above - I'm experiencing the same issue on Chrome 97.0.4692.99, Windows 10 OS. Various buttons and sliders' active areas don't match their visual elements. This extends to the 'fullscreen' button which, once fullscreen, is no longer clickable because the active area for the 'Report an Issue' link extends across the 'fullscreen' visual element.

hecksadecimal:
Seems like this isn't just an issue with fullscreen mode. 

Firefox 96.0.3 (64-bit) on Manjaro 21.2.2 (Kernel 5.15.12-1-MANJARO)
![image](https://user-images.githubusercontent.com/20251491/151877973-c0148e66-a8f9-4612-8ad5-f6a1e4a97c23.png)

johnnygossdev:
> Just adding to the above - I'm experiencing the same issue on Chrome 97.0.4692.99, Windows 10 OS. Various buttons and sliders' active areas don't match their visual elements. This extends to the 'fullscreen' button which, once fullscreen, is no longer clickable because the active area for the 'Report an Issue' link extends across the 'fullscreen' visual element.

This is my experience too on Firefox / Windows.

https://user-images.githubusercontent.com/3285214/151990012-b6706b0f-b41b-454f-aac9-befecbbabd67.mp4

[...]

Issue #253, opened 2022-02-04
Mouse misaligned with buttons in Firefox


AimVaguely:
**Describe the bug**
When hovering my mouse over the buttons on the top right (report, options and full screen) they only change colour and are clickable when my mouse is significantly to the right of them, this means that the report button only lights up when I hover over the full screen button

**To Reproduce**
Steps to reproduce the bug:
1. Open page with buttons in top right (e.g. https://gdquest.github.io/learn-gdscript/#course/lesson-5-your-first-function/lesson.tres)
2. Hover over report button
3. Try clicking
4. Nothing happens
5. Move mouse over full screen button
6. Click
7. Page opens to report

**Expected behavior**
The button is clickable when my mouse is directly above it.

**Screenshots**
![image](https://user-images.githubusercontent.com/94509482/152485395-212a7153-5150-40bf-a278-b3b123082d60.png)

**Information about your device (please complete the following information):**
 - Operating System: Linux x64 (Manjaro)
 - Browser: Firefox 96.0.3 (64-bit)

**Additional context**
This still happened when I opened the page in a private browser. This bug didn't happen when opening in chrome.

NathanLovato:
Could it be the same issue as #208? Depending on people's display and context, there seems to be an issue in Firefox between the visuals and interactions.

AimVaguely:
It might be, I do think I had a similar problem in some other places, like with selecting text in the text area in the practices, but I don't remember it being as big as a difference between where the mouse is shown and where it is highlighting elsewhere.

There might also be a difference with #208 because that seems to only be vertical difference, but I was talking about a horizontal difference, so that might make it a different cause. They do sound similar though, I just figured that they were different enough to make a new issue about it.

AimVaguely:
Just found out the buttons on the top left are misaligned the opposite way as the buttons on the top right. My guess is that the button is proportionate to the width of the screen, but the scaling with the black bars on the left and right side of the screen mess up where it is displaying it as.

I also tested making it full screen so that there are no black bars (it seems to be locked to an aspect ratio) and the buttons match up with the mouse now.

NathanLovato:
I see, then I think it's the same or a related issue.

@Xananax is this + #208 something you could look into? The app's interactivity gets offset from the UI components' visuals in Firefox.

NathanLovato:
Is this still happening with the latest staging version? You can test here: https://gdquest.github.io/learn-gdscript/staging/

[...]

PR #278, opened 2022-02-07
Attempt to correct fullscreen woes

Xananax:
This PR aims to replace the Godot-side full screen handling with a browser-side fullscreen handling.

This should partially or completely address:

- #272
- #226
- #208

## What Was The Challenge?

Two main problems:

1. There were two competing systems to set the app fullscreen or exit fullscreen. Godot can set the canvas context to fullscreen, the browser can set the window to fullscreen, and both can un-fullscreen the canvas context or the window. Neither system is aware of the other.
2. When setting the canvas to full screen, Godot's UI interactable area does not correspond to the visual UI on screen. This happens even if the canvas is set as full screen through Javascript. I do not know if the failure is on Godot's end, the browser's end, or due to our CSS (it'd still be a bug, but would explain why it isn't more widespread).
 
It seems that when setting the _document_ as full screen (as opposed to _the canvas context_), there's no UI problem.

Note that we cannot call the browser API to set fullscreen from Godot (for security reasons, browsers disable the functionality), so the button to set the document as full screen has to be an html button

This commit addresses problem 1 (#226) by:
- making sure the browser polls fullscreen changes and dispatches a signal to Godot
- making sure Godot can listen to the signal

It also addresses #272 by:
- removing the fullscreen button in the Godot app

Finally, it partially addresses problem 2 (#208) by:
- removing the Godot fullscreen button in browser contexts
- removing the fullscreen Godot shortcut in browser contexts
- adding an overlay html button in the top right corner

"Partially" because there's mention in #208 of UI becoming shifted from clickable areas _without_ any fullscreen happening. I have not been able to reproduce this.

## Thoughts For the Future:

1. We should/could explore the different combinations of Godot stretch modes and CSS, and see if we can fix this in a more idiomatic manner. This will take a long time and require a lot of time-consuming exports to HTML and testing in many browsers. I am not sure it's worth it.
2. At the moment, Godot-JS translations are distributed across several files. We could reduce code and possibly clarify intent if we moved all JS marshalling in a single autoload (possibly `Events.gd`?).


Xananax:
If we want a more long-term solution, we could consider leaving this PR in the fridge while I explore scaling issues. Alternatively, we could roll it out, then revert it once/if a proper scaling solution is found

NathanLovato:
If it improves the situation, I'd definitely ship and later consider more issues. We have a limited workforce so I'll take of course full fixes but also partial ones and other improvements.

NathanLovato:
In Brave it might've been a caching issue. I built once and tested both browsers in parallel.

Testing tonight, both in Brave and Firefox, now the button appears. Clicking the fullscreen button after loading the app causes the following error in both browsers:

```
Uncaught (in promise) TypeError: Failed to execute 'exitFullscreen' on 'Document': Document not active
    at HTMLButtonElement.toggle (bootstrap.js:408:20)
```

And fullscreen doesn't toggle, unfortunately.

Xananax:
After more investigation; I have _no documentation about this_, it's entirely conjecture, but it _seems_ that all browsers have 2 "fullscreen" mode, which are not doing the same thing at all. I'll call them `mode API` and `mode native`

## Mode API

The fullscreen triggered by the _Javascript API_ takes an element (normally, a video, such as a Youtube video) and fills the screen with it.

## Mode Native

The fullscreen mode triggered by <kbd>F11</kbd> _maximizes_ the browser chrome, removes bars (but leaves the sidebar if it is open). This mode is clearly different from `mode 1`. 

## Some key differences:

Mode API:
1. a notification appears on top saying "so and so is now full screen"
2. escape automatically escapes full screen

Mode Native:
1. an animation plays, where the bars slowly shrink out of view
2. the top address bar shows when the mouse hits to top part of the screen
3. escape does _not_ exit full screen, pressing <kbd>F11</kbd> again is mandatory

## Some similarities

_both modes_:
1. set the css mode to fullscreen, which means that my media query `@media all and (display-mode: fullscreen) {` unfortunately runs on both, and I can't differentiate them
2. Confusingly, you can do _both_ F11 _and_ press a maximize button
3. <kbd>F11</kbd> exits _both_  fullscreen modes.

## What _doesn't_ work:

1. hijacking `f11` to trigger our own fullscreen; browser security prevents it, it has to be a user pressing a button
2. doing `document.body.requestFullscreen` or any other combination; `window.requestFullscreen` doesn't exist

## Prior literature

This fullscreen issue isn't discussed anywhere, but I found references to the difficulty of fullscreen in general;

- [Google Web Discussion](https://developers.google.com/web/fundamentals/native-hardware/fullscreen)
- [A module by Sindre: Screenful](https://github.com/sindresorhus/screenfull)

Checking the [screenful demo](https://sindresorhus.com/screenfull/) or any youtube video will demonstrate the two different behaviors.

Pictures attached below

### Screen Full Normal, Mode API, Mode Native

screenfull normal

screenfull mode api

screenfull mode native

### Screen Full Normal, Mode API, Mode Native

youtube normal

youtube mode api
 
youtube mode native


## Mitigations:

1. remove our fullscreen button entirely, let the browser handle everything
2. disable <kbd>F11</kbd>. While we cannot _trigger_ fullscreen from a shortcut, we can simply prevent it from working
3. do it half-half, and hide the button _if the fullscreen was not triggered by a press_. This is detectable fairly reliably

NathanLovato:
What would you recommend? We need to find a way to solve the issue with the UI not matching where the user has to click, and ideally make full-screen toggles "just work" as far as the user is concerned. Whatever gets closest to fixing that seems good to me.

Xananax:
I'll try doing 2 & 3.

1. if user clicks, the user can click again. the user can also press escape to exit
2. if the user presses <kbd>F11</kbd>, then nothing happens
3. if the user opens the menu, and presses full screen from there, then the button hides. The user can then access the menu again to exit

Potential issues:

1. top buttons need to be moved to the left
2. someone could have binded `fullscreen` to a different key
2. someone could click the button, _then_ press in the black bars, _then_ trigger native fullscreen, which means they'd have to exit two full screen context. It's a stretch, but technically possible.

Other possibility that I didn't mention:

We could also investigate more fullscreen from Godot and why it isn't working. I did try a bit, and the same problems occurs with our stretchy CSS removed, so it seems our CSS is unrelated to the problem (but I could investigate more)

NathanLovato:
We could probably get in touch with Faless for extra insights and for insights about #253... I don't know if he'd have time but we could also hire him for that, as he's the person who should know the most about Godot's HTML5 exports. He could probably solve the issue much for efficiently than us.

Xananax:
If 2 & 3 don't work, then the cost efficient to deal with this is probably to simply remove the option of full screen completely. Someone wants to do that, they can use the browser functionality.

We could research this more, but the cost is starting to be disproportionate in regards to the yield

NathanLovato:
Definitely yes. I'm personally fine with the state of full-screen, but it's an issue if the interaction zones don't update for some users.

Note that some users reported this happened to them on displays with ratios that are *not* 16:9. Widescreens, ultrawide displays. If your display like this too?

NathanLovato:
It's working well now.

I found one edge case.

In brave, if you press F11 the button gets in focus but doesn't take you to fullscreen. If you then click it once while in focus, you get to fullscreen. Click the button again and you get out of fullscreen. After that, clicking the button or pressing F11 doesn't do anything anymore

This issue only occurs when first pressing F11 to make the button in focus.

In Firefox, pressing F11 focuses the button but doesn't toggle fullscreen. However, the issue you see on Brave doesn't occur: toggling fullscreen by clicking always works. I'll also reinstall chrome to start testing there systematically, as it's probably the most used browser by our students.

NathanLovato:
Same result as Brave when testing in Google chrome

Xananax:
I was able to repro once, but no more. It's very strange. The strangest part is I got this on recording because I wanted to show you I couldn't repro. So I have the exact steps, but despite trying them again, as well as a lot of variations, I couldn't reproduce it another time.

I tried nonetheless to guess what happened and I may have fixed it. If your can consistently repro, may I ask you to run the app with this url: http://localhost:8000/?debug=fullscreen

Do _not_ open the inspector (in case that plays a role in any sort of way), then, if you can reproduce the bug, _then_ open the inspector and send me what's written there.

PR #309, opened 2022-02-11
content: add lesson 17 introduction to while loops

### LINKED ISSUE #303: Lesson 17: Introduction to while loops

johnnygossdev:
- [ ] Create tests for practice 1

---

This PR adds lesson 17, which introduces while loops.

closes #303 

image

The tests for the practice are more complicated, so suggestions and implementations are welcome.

The desired solution from the user would be something like:
```
func move_to_bottom():
	while cell.y < board_size.y - 1:
		cell += Vector2(0, 1)
```

The issue is that the code can run a varying amount of times.


johnnygossdev:
The lesson is ready for a review @NathanLovato. There's also a practice there too, but the tests aren't implemented because I could do with a hand with them after what we discussed.

NathanLovato:
Okay I'll work on this later today

course/lesson-17-while-loops/lesson.tres
```diff
@@ -0,0 +1,237 @@
...
+[sub_resource type="Resource" id=1]
+script = ExtResource( 2 )
+content_id = "res://course/lesson-17-while-loops/content-DizKUdOC.tres"
+title = ""
+type = 0
+text = "Our new knowledge of vectors can help move the robot move in a grid.
+
+We can define the size of a board using a [code]Vector2[/code].
+
+Each square is a [i]cell[/i]."
```
NathanLovato, line 18: I think I would rather state the problem clearly and mention loops from the start. In short: we've seen how we can use functions to reuse code, now we'll see how to use loops to _repeat_ code.

And then go on to the explanation that we can use vectors to move things (note that I added info about that in l16 so l17 will reinforce the idea)

course/lesson-17-while-loops/lesson.tres
```diff
@@ -0,0 +1,237 @@
...
+visual_element_path = "visuals/ExampleBoard.tscn"
+reverse_blocks = false
+has_separator = false
+
+[sub_resource type="Resource" id=3]
+script = ExtResource( 2 )
+content_id = "res://course/lesson-17-while-loops/content-Q2XNwQMq.tres"
+title = ""
+type = 0
+text = "Let's add our robot onto the grid.
```
NathanLovato, line 38: I would also first show this example of the robot moving on a grid and then break down how the grid works. Basically I feel the start of the lesson is lacking context: the title is about loops, but you talk immediately about a grid and moving on a grid.

NathanLovato (commented):
On top of the two comments, the student is lacking some info about what while loops are for. They have some important uses, but I think in the current state and with the next lesson, people could feel while loops have no uses - as we discussed with Jad in the chat some time ago, all good uses of while loops tend not to be for beginners (reading a stream, buffer, parsing, etc.)

NathanLovato:
Another thing is you tend not to use many headings. There are natural cutting points that would benefit from being their own sections, like when you first introduce the while loop. It's mostly to help separate things in the user's mind

NathanLovato:
I just inserted this one for the sake of example

image

NathanLovato:
Last issue, whenever you can, you want to use Center in Frame in code examples - the lesson column can change size, in which case your grid will be off.

This means you need to design your example scenes so they draw around the scene's origin rather than towards the right and bottom.

johnnygossdev:
Thanks for the feedback, I'll check and change lesson 18 as necessary.

NathanLovato:
Just a small note, I left a comment about missing info about when to use while loops, but only saw later they were covered, just they were in a learn more box. So I moved them to main content, around the end of the lesson.

PR #312, opened 2022-02-12
feat: animate turtle turning

### LINKED ISSUE #5: Animate the turtle turning
Calls to turn_right() or turn_left() work but don’t make the turtle turn visually. Animating the turn on generated vertices would make it easier for students to understand what the function does (the turtle virtually turns in place but does not draw anything).

With this change, when the turtle reaches a vertex, it would turn in-place based on calls to turn_right() and turn_left().

For an example of the issue, see the runnable example in lesson 5 that only draws one line.

### LINKED ISSUE #108: Lesson 6 practice "Drawing rectangles of any size" solution has an extra turn_right(90)
After entering the solution, the rectangles are drawn correctly. However the output produces the following error:

"ERROR The drawn shapes don't have the expected length and height. Did you forget to use the length and height parameter?"

Screen Shot 2022-01-04 at 9 56 40 PM

Clicking the show solution button shows that an extra turn_right(90) is needed at the end of the draw_rectangle function, when it shouldn't be needed.

NathanLovato:
With this PR, the turtle will animate turning and jumping, which should make function calls like turn_right and turn_left more intuitive.

close #5 
fix #108


NathanLovato:
Left to do:

- Ensure drawing order works - lines can end up above labels
- Clean up labels on top of lines when resetting
- I'm thinking of just drawing outlined circles without the numbers for vertices, as we don't use the vertex index anywhere in the course - it's unnecessary info for the student I think.
- Figure out a way perhaps to move the camera to the middle of polygons that are about to be drawn. To do that, we can store commands in a temporary array until closing a polygon, then, when closing it, insert a command to move the camera before all the commands in the buffer.

PR #327, opened 2022-02-16
content: add lesson 18 introduction to for loops

johnnygossdev:
- [x] Write lesson
- [x] Add practices
---
This lesson shifts the focus from while loops to for loops. We refactor the previous while loop example to use a for loop instead.

It also explains the `range()` function and how we use it to run code a set amount of times.

image


NathanLovato:
I'm reviewing now. Quick note: please avoid writing explanations in code comments in this app, as that'll simplify localization work quite a bit

PR #345, opened 2022-02-20
content: add lesson 22 strings

### LINKED ISSUE #308: Lesson 21: Strings

johnnygossdev:
This PR adds lesson 22: Strings. It's a relatively short lesson which looks are what strings are and why we use them.

It has a few practices which use strings with other learnt concepts like arrays.

- [x] Write lesson
- [x] Add practices

closes #308

image

https://user-images.githubusercontent.com/3285214/155763313-4dbe2fcb-d002-450d-a5f0-7bd38e2a88dc.mp4


johnnygossdev:
This lesson is ready for review @NathanLovato, but I'm unable to merge `main` into this branch and resolve conflicts due to filenames which aren't compatable with Windows. Jad mentioned the `::` was the issue.

```
Error: Command failed: C:\Program Files\Git\cmd\git.exe -c core.longpaths=true -c core.quotepath=false -c color.ui=false checkout main --
error: invalid path 'course/lesson-19-creating-arrays/moving-turtle/MovingTurtle.live-editor/scripts/MovingTurtle.tscn::1.tres'
```

johnnygossdev:
Also I'd like to catch this error because missing `""` could be pretty common with strings.

image

But the app crashes instead.

image

johnnygossdev:
As a final note, it'd be good to ceate punching animations and swap the default animation with that, but I didn't have time to create them.

course/lesson-22-strings/lesson.tres
```diff
@@ -0,0 +1,198 @@
...
+visual_element_path = ""
+reverse_blocks = false
+has_separator = false
+
+[sub_resource type="Resource" id=3]
+script = ExtResource( 3 )
+content_id = "res://course/lesson-22-strings/content-DpPwQDwf.tres"
+title = "What are strings"
+type = 0
+text = "A string is a variable type which holds text instead of numbers. When we use strings, we use quotation marks (\"\") to differentiate them from other variable types, or method names for example.
```
NathanLovato, line 44: You shouldn't talk about variable type, at this stage, variables are dynamic. Also, String itself is a _value_ type, you don't need a variable to hold it.

course/lesson-22-strings/lesson.tres
```diff
@@ -0,0 +1,198 @@
...
+[ext_resource path="res://resources/ContentBlock.gd" type="Script" id=3]
+[ext_resource path="res://course/Documentation.tres" type="Resource" id=4]
+[ext_resource path="res://resources/Practice.gd" type="Script" id=5]
+
+[sub_resource type="Resource" id=1]
+script = ExtResource( 3 )
+content_id = "res://course/lesson-22-strings/content-jotxF5HU.tres"
+title = ""
+type = 0
+text = "Throughout this course, we've mostly stored numbers in variables. But what if we wanted to store a player's name? Using only numbers for names isn't very user-friendly at all, as 70801 here will tell you."
```
NathanLovato, line 14: Actually, this first example, which gives the robot a number as a name, uses a String. I'm unsure we should use that as a result, because as a dev you need to be really clear with types, their differences, and when you need to convert them (like here, you'd convert a number to a string to display it in a label)

course/lesson-22-strings/lesson.tres
```diff
@@ -0,0 +1,198 @@
...
+visual_element_path = "visuals/PrintString.tscn"
+reverse_blocks = false
+has_separator = false
+
+[sub_resource type="Resource" id=7]
+script = ExtResource( 3 )
+content_id = "res://course/lesson-22-strings/content-5tT6n1Tx.tres"
+title = ""
+type = 0
+text = "Strings are like an array of characters. In fact, we can even loop through the characters of a string in the same way."
```
NathanLovato, line 66: Not _like_, they are arrays of characters. Just in GDScript, it's happening under the hood as there's no exposed character type.

course/lesson-22-strings/lesson.tres
```diff
@@ -0,0 +1,198 @@
...
+has_separator = false
+
+[sub_resource type="Resource" id=11]
+script = ExtResource( 3 )
+content_id = "res://course/lesson-22-strings/content-6fUqbTxY.tres"
+title = ""
+type = 0
+text = "Every piece of text you see in this app is a string that Godot is displaying for us.
+
+Much like how [code]Vector2[/code] variables make calculations easier, string variables come with lots of helper methods and tricks we can use.
```
NathanLovato, line 124: We've used the term function throughout the app so far. We're not gonna cover OOP or objects so we can't really talk about methods and properties, which are OO concepts

NathanLovato:
Just left you a few comments for future updates

> As a final note, it'd be good to ceate punching animations and swap the default animation with that, but I didn't have time to create them.

There's no time limit on this. Can you add them? The combo practice doesn't feel like a combo without them. If not let me know I'll take care of those

johnnygossdev:
Great, thanks. Sure, I can add the animations and perhaps remove the first visual with the number name. I thought it might be misleading when I added it, but wasn't sure how else to add a visual to it. Perhaps it's not needed.

NathanLovato:
I think it's fine with mostly code/print examples. It's a short lesson so it's not so long it'd get boring for the reader.

NathanLovato:
Hey Johnny, is this good to test and merge now?

johnnygossdev:
Hello!

I'd say so. I added different animations and updated the practice. There's
perhaps removing the first potentially confusing visual in the lesson with
the number. Otherwise good to go

PR #359, opened 2022-02-24
Initial support for german language

SpeCter:
We added temporary support for the german language.

Lesson 1 - 8 are mostly translated and only need small changes
Lesson 9 - 20 are only machine translated and need another pass

application, classref_database and error_database are mostly translated but some strings don't seem to get used by the app.


NathanLovato:
Would you mind copying the files over to our new repository, https://github.com/GDQuest/learn-gdscript-translations? I'm setting it up to support weblate.

SpeCter:
Sure! Will do that when I come home from work.

NathanLovato:
Thanks! I'll close this PR and let you move the files to the other repo then. Thanks much for your help.

PR #362, opened 2022-02-25
improvement: set cursor position in course builder

### LINKED ISSUE #326: Make cursor start where the student should start writing code in practices
The cursor in the SliceEditor always starts at the first character.

It'd be nice to either start at the last character or wherever we want the student to start writing code.

Current behavior:

image

Desired:

image

razcore-rad:
closes #326


NathanLovato:
In some cases, it doesn't work.

In editor, lesson 7 practice 1:

image

When testing the practice:

image

I'll finish updating the cursor position in practices real quick and push the updated practice resources.

NathanLovato:
Actually, there's an issue with saving, when I change the property manually it doesn't get saved. Changing either the line or column should mark the resource as dirty and trigger a save upon clicking the save icon or testing the scene.

NathanLovato:
Done going through all lessons. I only tested a handful, only lesson 7 practice 1 wasn't working, didn't give it a lot more attention asI have to finish other tasks. It might just have to do with the saving issue, not sure.

NathanLovato:
It was because you were setting the column before the line - so the cursor would get limited to the length of the first line. Fixed this, gonna rebase and merge now

razcore-rad:
That makes a lot of sense

PR #363, opened 2022-02-26
Add 5 lessons

### LINKED ISSUE #307: Lesson 22: Adding and removing elements in arrays

### LINKED ISSUE #354: Lesson 23: Accessing values in arrays
This lesson can come right before dictionaries because while you can find an index in an array, dictionary lookups can be easier and there will often be more efficient, especially when arrays get large.

While those performance details won't matter too much to students at this point, the dictionary is really much easier to understand and use than arrays for accessing specify keys than arrays.

The examples here could build upon the previous lesson and have us directly access the 3rd order, the for folder, et cetera in the queue without having to use a for loop or some other approach.

### LINKED ISSUE #355: Lesson 28: functions that return a value
We haven't covered making pure functions or functions that return values at this stage because the student had too few techniques in their toolbelt.

At this stage, we can make them write useful pure functions, for example, to filter arrays, or to convert vector coordinates from grid to screen coordinates and vice versa.

This lesson should also talk about specifying the return type of a function with the `->` syntax as we didn't get to cover it in the type hints lesson.

There are 3 advantages of functions that only calculate and return values:

1. They make your code easier to read.
2. You can test them in isolation.
3. As a result, they are easy to debug.

### LINKED ISSUE #356: Lesson 27: type hints
After discovering value types and seeing that we need to be careful with them, we introduce type hints.

Type hints are most useful when getting frequent compilation and linting in Godot so the scope of this lesson is a bit limited. Yet, writing types is a good learning practice: too many self-taught programmers rely on dynamic code and get stuck when a type-related error occurs.

This lesson should show how to specify the expected type of a variable or function argument. We haven't talked about functions that return values yet so we'll introduce the `->` syntax for return types in the corresponding lesson.

We could also introduce type inference here as a shortcut.

Finally, we should make the advantages of using strong typing or dynamic code clear:

- It is good for learning as you will get lots of type warnings that will help you improve your understanding of what the code is doing.
- Inside Godot, it will also improve auto completion: the computer will better understand your code and will be able to suggest code completions for you more efficiently.

### LINKED ISSUE #357: Lesson 26: Value types
We already talked about value types in the course up to this point, but users have only been writing dynamic code.

This lesson should just focus on value types. It should highlight the problems you will face if you do not master the types and don't understand them.

For example, when trying to add a number to a string directly, you will get an error. You need to convert the number into a string using the `str()` function.

NathanLovato:
Close #307
Close #354 
Close #357 
Close #356 
Close #355


NathanLovato:
@Xananax just so you know, we currently don't write code comments in English because of localization. We don't have a way to translate them atm.

NathanLovato:
@Xananax please let me know when you're done, I'll take care of conflicts and merge

Xananax:
@GDQuest/core The exercises are ready for review and testing. There's still tightening up to be done from the content point of view (exercise description, hints, graphics maybe)

NathanLovato:
Just tested, it's good as is I think. Just the second practice with the tracks being tight and the offset on the final chunks being small, it's a bit hard to read. I would avoid making the final part in U with no space between the tracks, and also offset the offset chunks more.

NathanLovato:
It's this part I find a little too busy:

image

Regarding graphics and polish don't worry, we'll get people to test the app and we'll have to make improvements based on that.

I'll make minor tweaks then solve the conflicts. I'll let you review what you want to, then please let me know when it's good to merge.

PR #372, opened 2022-02-28
fix(lesson10): Pick correct node to check in validation script

razcore-rad:
`_scene_root_viewport.get_child(0).get_child(0)` is returning the camera node so the validation scripts wouldn't work properly. This PR checks the correct node by retrieving it with `get_node()`


NathanLovato:
Tested, works well, thanks

PR #417, opened 2022-03-04
fix: prefer fuzzy font on scale rather than pixelated

razcore-rad:
I turned on `Use Mipmaps` and `Use Filter` on all fonts and it does improve a little bit although there still are corner cases. I think overall it's better than having it pixelated though it introduces a bit of fuzziness.

This is how first L23 exercise looks in the browser scaled by the view of the browser window (so it isn't fullscreen)
image

Compared to fullscreen:
image

I opened the PR to have a discussion if it's necessary to change the font specifically for some examples as well.

### DIFF
--- a/ui/theme/fonts/font_button.tres
+++ b/ui/theme/fonts/font_button.tres
@@ -4,4 +4,6 @@
 
 [resource]
 size = 24
+use_mipmaps = true
+use_filter = true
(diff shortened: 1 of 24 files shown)


razcore-rad:
The Godot documentation on `Use Filter` font setting recommends turning mipmap and filter on if we don't want a pixel game.

PR #432, opened 2022-03-07
improve: start in fullscreen mode by default

### LINKED ISSUE #400: Save window position, size and fullscreen state for subsequent runs on desktop

razcore-rad:
Reason I made this a draft is to confirm if the state should be saved and restored without user input. The state being: window size, window position and window fullscreen state. The current behavior is to save these to the user profile and restore it on subsequent runs.

If this is fine, only thing I'd do is add a reset button to the Settings panel to reset the state to the project default settings, in case someone wants to recenter the window or what-not.

closes #400

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -397,6 +397,7 @@ gdscript/warnings/unsafe_call_argument=true
 
 window/size/width=1920
 window/size/height=1080
+window/size/fullscreen=true
 window/size/test_width=1080
 window/size/test_height=600
 window/dpi/allow_hidpi=true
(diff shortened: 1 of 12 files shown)


razcore-rad:
Alright so I implemented the functionality based on the last discussion: starts in fullscreen mode by default, added quit buttons (in welcome screen, end screen and top bar).

NathanLovato (changes requested):
You gotta check the web version as there, the fullscreen button is not part of the app

image

PR #437, opened 2022-03-08
improvement(lesson15): rethink color coding and box alignment

### LINKED ISSUE #430: Color "Confusion" in Modulo lesson 15
**Issue description:**
In Lesson15 Modulo in the modulo illustration on the right. 
(Which i really like 😊 )

The "number" slider is green and show by green boxes. Makes sense.
The "modulo" is grey but the grey boxes illustrate the remainder. 

image

_Suggestion:_
Move the math part with "Remainder: x" under the sliders. 
Change color of "Remainder: x" test to reflect the color of the boxes. 
Change the colors for Modulo slider to one that is not used on the boxes.

razcore-rad:
The example now looks like this:

https://user-images.githubusercontent.com/1177508/157216983-71bb1f6f-3e49-469f-a0ab-898ce4bb311b.mp4

To color-code the sliders I had to add a feature to  `RunnableCodeExample.create_slider_for()` to pass an optional color. This color overrides the color of the labels and the slider stylebox. It will also force only the highlight stylebox be visible at all times, even if the mouse doesn't hover over it.

closes #430

### DIFF
--- a/course/lesson-15-modulo/visuals/ExampleModulo.tscn
+++ b/course/lesson-15-modulo/visuals/ExampleModulo.tscn
@@ -7,27 +7,16 @@
 script/source = "tool
 extends RunnableCodeExample
 
+
 func _ready():
 	yield(self, \"ready\")
-	var number_slider = create_slider_for(\"number\", 1, 7, 1)
-	var modulo_slider = create_slider_for(\"modulo\", 1, 7, 1)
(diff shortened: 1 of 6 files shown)


razcore-rad:
Ah my bad, I shouldn't have commited `course-learn-gdscript.tres`. Let me fix that quick

PR #469, opened 2022-03-25
Add offline script verifying support using custom build of Godot

### LINKED ISSUE #300: Remove the need for a server to check code errors
After talking with core contributors vnen and Faless, it seems we could get the GDScript compiler running in the browser and use that to check for user errors.

If we could get rid of the server entirely, that'd greatly improve the user experience - no more stalls, timeouts, or need for an internet connection. That'd allow us to distribute the app on desktop more easily too.

This task is about exploring this trail. I asked Faless if he'd have time to help, let's see his reply before working on this ourselves, as he would have a lot more experience with this.

This feature would also fix #310 and any connection issue, like the ones presented in #462.

### LINKED ISSUE #310: Godot will not reconnect to server if OS connection happens after app is launched
If you launch the app with wifi off, the app is disconnected, as it should. If you then connect, Godot can't reconnect to the given hostname.

The error seem to indicate that Godot tries to resolve the hostname once, then caches that and doesn't try again.

image

This is a function that tries to reconnect every 3 seconds in case of failure. I was disconnected for the first 2 ~ 3 lines, but I was connected after that

**To Reproduce**
Steps to reproduce the bug:

1. Either set `ConnectionChecker`'s `print_debug_strings` to `true` (or observe the connection light in any Practice)
2. disable your OS connection
2. launch the app
3. Note that the connection light in Practices says you're offline, and ConnectionChecker prints "attempting to connect..." repeatedly
4. Note that the `Errors` tab contains `connect_to_host: Condition "ip_candidates.empty()"`
5. Turn your connection OS on
5. Despite [`WebSocketClient.connect_to_url`](https://github.com/GDQuest/learn-gdscript/blob/01b64556d0410a70c27e9ebe5214887b3b40f010/autoload/ConnectionChecker.gd#L65) being called again, and the server being available, Godot won't connect 

**Expected behavior**

Godot should try to resolve the hostname again (if the error is correct and if this is the problem)

 - Operating System: Linux

Razoric480:
- This refactors the code to split the `ScriptVerifier` class to make it into a base class, with `test()` as a virtual function to override, and implements a `OfflineScriptVerifier` class. Common functions and data have been moved into the `ScriptVerifier`.
  - LSP based verification has been removed.

- The `OfflineScriptVerifier` class depends on the `GDScriptParserWrap` Reference class, which is not part of Godot by normal. It requires a custom build of Godot. I've uploaded export templates that include this class at https://github.com/Razoric480/godot/releases/tag/3.4.4-parser, which are compiled on my fork https://github.com/Razoric480/godot/tree/raz/gdparser-expose. Including them as a binary data through a git push is a bad idea given the size. I'm unsure of a better way - would be some CI magic of sorts I imagine.

The `GDScriptParserWrap` class exposes the `parse(script_content: String)`, `has_error()`, `get_error()`, `get_error_line()` and `get_error_column()` functions by wrapping Godot's `GDScriptParser` class, which isn't normally visible through scripting. I didn't expose warnings because they are only visible during tools (though we blacklist anything of that severity anyway.)

- `global/...` settings removed.
- References to the LSP mostly removed, files and classes renamed in consequence, like LanguageServerError just being ScriptError and LanguageServerRange becoming ErrorRange.
- References and UI related to connecting to a server removed.

Close #300 
Fix #310


NathanLovato:
Thanks for your work @Razoric480, this is a massive improvement.

May I ask you to remove the option to switch entirely, and rather remove the LSP part? It'll still be in the code history should we ever need it (we could keep it in a branch), but the idea's to get rid of it entirely. I can help with removing UI bits like the server connection indicator.

Razoric480:
@NathanLovato Done - I've excised 99-ish% of LSP-ness from the project. Preliminary test reveals it's still working, though users have a knack for finding issues so we'll have to see if I messed something up.

NathanLovato:
Thank you very much François, great work you did there, as usual :)

I'm a bit busy preparing the course release but will test and merge as soon as I'm through with my todo.

NathanLovato:
After testing, it's working like a charm. I'll address some bugs in the app today, make a stable release, then we can merge this and update the CI for the following release.

Issue #448, opened 2022-03-12
Warning, the app is not suitable on mobile device, is not fully readable


jacky-brownsteen:
**Describe the bug**
I'm not sure if this is a bug or merely a text problem. It's for sure a veeery small one with low priority!
If I follow the link https://gdquest.github.io/learn-gdscript/ on my smartphone (/e/OS, Brave Browser), there pops up a warning, that the app is not suitable on a mobile device. So far so good. The only "problem" is, that the text is not fully readable as the text window is sized in a way, I cannot access all the text. Sometimes I was able to scroll to the right, but I never was able to see the text which is cut off on the left side.

**To Reproduce**
Steps to reproduce the bug:
1. Go to https://gdquest.github.io/learn-gdscript/ from a smartphone (/e/OS, Brave Browser)
2. See warning popup

**Expected behavior**
The full text should be readable.

**Screenshots**
![Screenshot_20220312-115854_Brave](https://user-images.githubusercontent.com/63466503/158015563-b2e17881-1f04-4509-849c-b860be06a051.png)

**Information about your device (please complete the following information):**
 - Operating System: /e/OS
 - Browser: Brave

NathanLovato:
We should definitely make the modal fit the browser and the text wrap as needed. Thanks!

PR #545, opened 2022-06-09
fix: detect mobile browsers specifically by name instead of using the browser size

### LINKED ISSUE #532: mobile browser wrongly detected
**Describe the bug**

I get the message:

>You appear to be using a mobile browser. This application is not suitable for small screens and portrait orientation. If you want to continue, it's at your own risk!

But i'm using Google Chrome on desktop Linux.

**To Reproduce**
Steps to reproduce the bug:
1. Go to https://gdquest.github.io/learn-gdscript/ (also happens on https://gdquest.github.io/learn-gdscript/staging/)
4. See message

**Expected behavior**
Desktop Browser should not be detected as Mobile Browser.

**Screenshots**

Screenshot from 2022-06-03 21-29-15

 - Operating System: NixOS 22.05beta67.6efc186e607 with Pantheon Desktop
 - Browser: Google Chrome 101.0.4951.64

**Additional context**
I have checked that i don't send a custom user-agent.

Feel free to give me further instructions to help debug this.

AnonymousXC:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #532 

This PR uses *javascript* to fix #532

### DIFF
--- a/html_export/static/style.css
+++ b/html_export/static/style.css
@@ -181,9 +181,9 @@ body {
 }
 
 @media (orientation: portrait), (max-aspect-ratio: 4/3), (max-width: 750px) {
-  .mobile-only {
+  /* .mobile-only {
     display: flex;
-  }
+  } */
 
   .force-mobile .desktop-only {
     display: flex;
--- a/html_export/static/style.css
+++ b/html_export/static/style.css
@@ -181,9 +181,6 @@ body {
 }
 
 @media (orientation: portrait), (max-aspect-ratio: 4/3), (max-width: 750px) {
-  /* .mobile-only {
-    display: flex;
-  } */
 
   .force-mobile .desktop-only {
     display: flex;
--- a/html_export/index_template.html
+++ b/html_export/index_template.html
@@ -56,6 +56,12 @@
     <div id="version">branch and commit hash</div>
     <script type="text/javascript" src="$GODOT_URL"></script>
     <script type="text/javascript">
+
+      const IS_MOBILE = /iPhone|iPad|iPod|Android/i.test(navigator.userAgent);
+      if (IS_MOBILE) {
+        document.querySelector(".mobile-only").style.display = "flex";
+      }
+
       const GODOT_CONFIG = $GODOT_CONFIG;
       const GDQUEST_ENVIRONMENT = {};
     </script>


NathanLovato:
@Xananax could you review this one?

html_export/static/style.css
```diff
@@ -181,9 +181,9 @@ body {
 }
 
 @media (orientation: portrait), (max-aspect-ratio: 4/3), (max-width: 750px) {
-  .mobile-only {
+  /* .mobile-only {
     display: flex;
-  }
+  } */
```
NathanLovato, line 186: That shouldn't be just commented out. If the class is going away from the css, then you should remove the lines.

AnonymousXC: Oh.


> Thanks for the contributions!

My pleasure

NathanLovato:
Okay from your explanation clearly we should drop this patch and do things differently. We should also address issue #448 along the way (warning popup can get cropped on some devices)

NathanLovato:
Thanks for the contribution but it appears it's not the way to solve the issue and so @Xananax will directly fix it shortly

AnonymousXC:
> Thanks! But I don't think this is necessary. I'll explain why:
> 
> There are 3 main concerns with using string-based device detection:
> 
>     1. You will get the warning even on large tablets, which could perfectly well support the application.
> 
>     2. You won't get the warning if you're on a small screen that isn't a mobile. You could be using a laptop with a small screen, or a portrait-oriented desktop screen, or resize your browser in such a way that the browser window is too small to support the application. There are also other mobile devices with small screens that aren't android or IOS (small laptops, small screens, etc. I myself own two Linux based mobile devices)
> 
>     3. A js-based detection on page load would not refresh later, whereas CSS would. For example, if someone was to access the application on a large enough tablet in portrait mode, rotating it would remove the warning, because CSS gets reeavluated. But with js-based detection, the warning would remain (unless we attach a listener to resize and a bunch of other events, and debounce this to not slow down the app).
> 
> 
> For those reasons, using userAgent is generally not advised and regarded as bad practice; it's not future-proof.
> 
> In our case, since we prevent usage instead of just changing the interface, it's even a little bit less advised than usual.
> 
> With this in mind, if we were use userAgent detection, here are the changes I'd envision to this PR:
> 
>     1. Add a class to body rather than add a style directly to the element. This allows styles to remain confined to CSS instead of defined in multiple places
> 
> 
> ```js
> if (IS_MOBILE) {
>   document.body.classList.add("is_mobile")
> }
> ```
> 
> ```css
> .is_mobile .mobile-only {
>   display: flex;
> }
> ```
> 
>     2. move the detection code inside the javascript file so the HTML file stays clean of any working code apart from the obligatory Godot settings. Otherwise, it'd be easy to forget this exists and not understand where the behavior comes from.
> 
> 
> But concretely, I don't think there is a bug to fix there.
> 
> The app effectively doesn't work well on sizes smaller than 750px, so warning users is the correct behavior, regardless of device. The warning is dismissable, so even in the case of false positives, the user can simply dismiss the message.
> 
> I'd suggest changing the message to not imply "mobile", that's all.
> 
> An additional tweak to the CSS can include:
> 
>     1. remove the `(max-aspect-ratio: 4/3)`, I suppose that's what triggering the message in the related bug
> 
>     2. lower the `max-width`, in case we can (I think it's the lowest possible value for which the app is well displayed, but we can verify this)

COOL.

PR #576, opened 2022-06-30
improvement: better wording in lesson 3

pcamp:


NathanLovato:
Thanks for contributing, but please don't make tiny changes in phrasing like this unless the sentences are really confusing or unclear. Small changes like these require changing all the translations (10+ languages in progress), so there should be a real need for changes.

Issue #315, opened 2022-02-13
When maximizing a Firefox window, the full-screen button disappears


NathanLovato:
@Xananax may I let you look into this? It happens only on Firefox. When the browser window is maximized, the fullscreen toggle button disappears. It reappears as soon as it's not maximized.

Tested on Linux.

https://gdquest.github.io/learn-gdscript/staging/

Dinock:
Maybe this is related to the bug:

When I switch to fullscreen on my external monitor the fullscreen toggle button gets pushed to the upper right corner.
Screenshot:
![16_10-firefox](https://user-images.githubusercontent.com/54438803/153773255-fbda6b74-1b35-459f-8694-3c185e140541.png)

This happens at 1920x1200 (16:10) on Ubuntu 21.10 with Firefox 97.0.

Xananax:
Explanation of the different issues:

1 - button disappearing when maximized:

The button disappears in fullscreen mode. That is because we cannot control the Browser's fullscreen, so we better bail out.

Browsers give you no way to know if they are fullscreen or not. The only way to check is through a series of hacks. One of the ideas I've had was to check for window width, and compare with document width. If they were close, within 2 pixels, I considered this "full screen". It worked well for me, but I guess it depends on your theme/browser/desktop.

I removed this. Only one hack remains now to check fullscreen status.

2 - button in the top right corner

This was expected, because the button is overlaid on top of the Godot app. The reason is that fullscreen from inside Godot was messing up, so we had to use a different method.

I now made it so the button tracks the position of the Godot game, giving the illusion of being part of the game (with some de-syncing during resizes)

3 - button disappearing if browser in fullscreen in another tab

As I said earlier, Godot cannot control browser fullscreen. They're two different modes. So if the browser goes to fullscreen, we hide the button as to not mislead the user into thinking the button can have an effect.

NathanLovato:
Tested in Chrome, Brave, and Firefox, it's now working for me. Thanks!

Dinock:
As I remember Xananax writing in another GitHub-issue that there is almost zero documentation about the buggy fullscreen-mode in browsers, wouldn't it make sense to reach out to the devs of Firefox and Godot to heighten their awareness of this issue? 
The GDQuest-App will most likely not be the last project which will encounter these kind of problems.

Xananax:
I expressed myself in a misleading manner. It isn't "a bug"; it's "working as intended". You're not supposed to allow random sites to:

[...]

Issue #381, opened 2022-02-28
Text is blurry in HTML5 exports


NathanLovato:
The text appears blurry at 100% browser zoom levels since recently. It might be since fixing scaling issues in the browser.

This is at 100% page zoom level in Brave:

![image](https://user-images.githubusercontent.com/12694995/156050862-9cf6764a-9296-45fb-bb0f-fe8d64a8aefd.png)

To get sharper text, I have to zoom back to 80% page zoom level:

![image](https://user-images.githubusercontent.com/12694995/156050949-9c473404-223c-4307-99a9-ce970b18d870.png)

The second pic is how the text should look like.

@Xananax any thoughts what might be causing this? The issue is worse on itch.io or when the page is scaled down/not full-screen.

iulover99:
This is also happen in Microsoft Edge too

PR #585, opened 2022-07-12
WIP: Blurry fonts

Xananax:
This is related to #381 

## Hypothesis

Because Godot uses rendered fonts, they do not resize as sharply as vector fonts. This is especially apparent if the viewport is not a multiple of the original resolution (1920x1080).

## Fix

The current PR makes sure the viewport resizes to multiples of `240px`. That is, 240, 480, 720, 960, etc. The hope is that pixel division on even numbers will yield sharper results.

This seems to work well enough for me, but it requires more testing.

In parallel, I asked itch to give me access to CSS editing there, so the same setup can be reproduced on the itch page.

Importantly: **we cannot ensure** sharpness if people have custom zoom values set up. A zoom of 72% will always look bad, no matter what. A scale of 125% will similarly result in a pixel division of 0.625, which will create aliasing on the edges.

If we wanted to go around that, we'd need:

1. to pass the current window size to Godot through the javascript API
2. have Godot use a different font size dependent on passed the window size
3. have Godot lock the size in increments of 240 too (set the mode to not stretch, and handle viewport changes ourselves)

This is also what we would need to do if we want to handle blurriness on Desktops.

It's possible, but I don't know if it is worth it at the moment. Godot 4 has automatic font scaling, so the problem could potentially be less visible once we port.

### DIFF
--- a/html_export/static/bootstrap.js
+++ b/html_export/static/bootstrap.js
@@ -68,7 +68,9 @@ window.GDQUEST = ((/** @type {GDQuestLib} */ GDQUEST) => {
       if (!isDebugMode) {
         return { app: true };
       }
-      const modulesList = params.get("debug").split(",").filter(Boolean);
+      const modulesList = (params.get("debug") || "")
+        .split(",")
+        .filter(Boolean);
       if (modulesList.length == 0) {
         return { "*": true };
       }
@@ -186,22 +188,23 @@ window.GDQUEST = ((/** @type {GDQuestLib} */ GDQUEST) => {
       console.error(msg);
       setStatusMode(StatusMode.NOTICE);
       const statusNotice = document.getElementById("notices");
-      msg.split("\n").forEach((line) => {
-        statusNotice.appendChild(document.createTextNode(line));
-        statusNotice.appendChild(document.createElement("br"));
-      });
+      statusNotice &&
+        msg.split("\n").forEach((line) => {
+          statusNotice.appendChild(document.createTextNode(line));
+          statusNotice.appendChild(document.createElement("br"));
+        });
       is_done = true;
     };
 
+    const loaderElement = document.getElementById("loader");
     /**
      * Grows the visual loading bar
      * @param {number} percentage
      * @returns
      */
     const displayPercentage = (percentage = 0) =>
-      document
-        .getElementById("loader")
-        .style.setProperty("--progress", percentage * 100 + "%");
+      loaderElement &&
+      loaderElement.style.setProperty("--progress", percentage * 100 + "%");
 
     /**
      * Callback used during the loading of the engine and packages
@@ -289,9 +292,12 @@ window.GDQUEST = ((/** @type {GDQuestLib} */ GDQUEST) => {
       localStorage.setItem(KEY, "true");
     };
 
-    document
-      .getElementById("mobile-warning-dismiss-button")
-      .addEventListener("click", forceAppOnMobile);
+    const mobileWarningButton = document.getElementById(
+      "mobile-warning-dismiss-button"
+    );
+
+    mobileWarningButton &&
+      mobileWarningButton.addEventListener("click", forceAppOnMobile);
 
     const currentValue = JSON.parse(localStorage.getItem(KEY) || "false");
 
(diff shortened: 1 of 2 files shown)


NathanLovato:
This doesn't help, but testing in the browser, I see the issue: the resolution of the <canvas> element changes with the app. The problem is that when the resolution goes down, the canvas just lacks resolution, and it's normal that the text becomes blurry.

I can get crisp text by forcing the canvas to stick to a resolution of 1080p and giving it the styles height:100% and width:100%, so it doesn't overflow from its parent frame.

image

And so, I think it's just a matter of changing the canvas-container so it doesn't lower the resolution of the child canvas too much.

I noticed this code in the JS

```js
    const onResize = () => {
      const { width, height, ratio } = snappedAspectRation();
      canvas.width = width;
      canvas.height = height;
      console.log({ width, height, ratio });
      canvasContainer.style.setProperty(`width`, `${width}px`);
      canvasContainer.style.setProperty(`height`, `${height}px`);
      document.documentElement.style.setProperty("--scale", `${ratio}`);
    };
```

It seems like the cause to me.

NathanLovato:
Current result: 

image

With a fixed canvas resolution of 1080p

image

Xananax:
It used to be set at 100%, the javascript code that resizes the canvas was made specifically to answer another bug... I'll try digging to remember

Xananax:
It was that commit: https://github.com/GDQuest/learn-gdscript/commit/ca3d17fe639816c9d20b79dac240df9f6a12071e

It fixes bug #253 (the commit also fixes #315 incidentally, but that's not important for our purposes)

So it seems we have the following options:

- 100% width and height, but interactive areas don't align with the graphics on Firefox.

or

- width and height set in JS, but then the canvas lacks enough pixels to display fonts properly on zoomed-in devices.

Another option is to 

- let Godot handle width and height, as it does in the default template, but we didn't want that because... Of some yet another bug. I'll go hunting.

NathanLovato:
Ah that was why!

Xananax:
The reason why we aren't letting Godot handle the size are:

- #226
- #272
- #208

 All three were fixed in https://github.com/GDQuest/learn-gdscript/pull/278

To summarize the reasons for that PR:

1. our fullscreen button in Godot could not know if the browser was set to full screen through <kbd>F11</kbd>, so the fullscreen icon could indicate the wrong state. So we decided to handle this in JS instead
2. we cannot call the JS to fullscreen from Godot, for sandboxing reasons. So we overlay a button in JS over the Godot frame
3. To overlay the button correctly, we need to input specific measurements, because some browsers (looking at you, firefox) don't allow us to query back the canvas size
4. Therefore, we decided to set the canvas at specific sizes, so the button could always be aligned (https://github.com/GDQuest/learn-gdscript/commit/ca3d17fe639816c9d20b79dac240df9f6a12071e)

In summary, our options:

1. Use Godot fullscreen:
  - <kbd>+</kbd>: less code, interactive overlays should work in all circumstances
  - <kbd>-</kbd>: cannot obtain browser state to change button style; cannot know if user use browser fullscreen
2. Use JS fullscreen + specific canvas sizes
  - <kbd>+</kbd>: button can be overlayed precisely, fullscreen state is tracked
  - <kbd>-</kbd>: font become blurry if the pixel aspect is not 1:1
 3. Use JS fullscreen + CSS 100% + query canvas size on size change
   - <kbd>+</kbd>: fullscreen state is tracked, fonts are not blurry
  - <kbd>-</kbd>: on firefox, querying is imprecise, button will not be placed correctly

I will try to set specific percentages in JS, a mix of 2 and 3, and see what happens.

Another option would be to overlay the fullscreen button at the bottom or top right, regardless of canvas size.

Xananax:
I allowed the canvas container (a simple div) to resize to specific sizes, but made the canvas fill the entire available space with `width: 100%; height: 100%`.

It's difficult for me to test (my screen is small and 1080p), but I think it's fixed? Functionally, everything works still (interactive areas, fullscreen button alignment, and functionality) on both Firefox and Chromium.

NathanLovato:
Just tested in firefox and brave, looking great!

PR #624, opened 2022-08-19
test: Add bare-bones integration tests for practices

razcore-rad:
Added a GUT test that iterates over the practices and runs the given solution.

There are some tests that fail and it might be due to how the `UIPractice._practece_completed` flag is set. For now I'm emitting a `test_student_code_completed` signal at the end of `UIPractice.test_student_code()`.

Other tests, like `res://course/lesson-7-member-variables/draw_rectangle_at_position/TestsDrawingRectangles.gd` fails to run and crashes GUT probably because it needs the scene to set up some variable

image

This needs further investigation.

The script is located at `res://gut/integration/test_course.gd`.


NathanLovato:
I just re-tested the PR, and many code solutions fail, even though they correctly work when running a practice as a user. Example: lesson 3's practices.

Could you look into this?

razcore-rad:
Should be fixed now

Issue #580, opened 2022-07-01
Give lesson exercises numbers or other quick identifiers


clockworkprince:
**Is your feature request related to a problem? Please describe.**
Right now, lesson exercises only have titles. This isn't the end of the world, but it would be easier to communicate faster about which exercise I'm talking about if there was a quick identifying character like e.g. a number or letter. In a list of items I sometimes have trouble processing which is the first, second, third, etc. entry, unless there are quick identifying characters to help me out. 

**Describe the solution you'd like**
I want the format of exercise titles to be something like "1. Exercise Title" or "A. Exercise Title" instead of just "Exercise Title".

**Describe alternatives you've considered**
I don't know if there's any alternative options. Maybe emojis, I suppose?

**Additional context**
Screenshot of the exercise choosing section at the bottom of a lesson right now: 
![GDscript From Zero - Request - Exercise identifiers A](https://user-images.githubusercontent.com/47702744/176883535-3afc8fe2-53be-4da8-b8e2-2fc22e0c9551.PNG)

My proposed edit: 
![GDscript From Zero - Request - Exercise identifiers B](https://user-images.githubusercontent.com/47702744/176883565-fbcfd6d8-dd6a-40c4-a712-704daa7c4176.PNG)

PR #630, opened 2022-08-23
Add practice index to lesson practices list and breadcrumb

henriquelalves:
- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #580 

Add practice index to the Lesson Practices list and on the breadcrumb. It doesn't add to the title of the `UIPractice` screen to follow consistency with `UILesson` title.


henriquelalves:
A quick note: at first I solve this by adding a "practice_index" to the Practice Resource and updating it when saving the Course (on the Course Builder), but:
1. The Lesson resource doesn't have a "lesson_index" field, so this would be inconsistent to the other resource.
2. The places where the Practice index was needed could just iterate using the Lesson it's contained in.
So I opted for the simpler solution. Lemme know if this was the correct take!

Issue #609, opened 2022-08-08
Change the end screen depending on the URL


NathanLovato:
We offered to integrate the app in the official Godot docs, but there shouldn't be our sponsored end screen there. We should have a different, neutral end screen to invite people to continue with the docs.

This task is about storing a flag in the user's profile if they do the app on docs.godotengine.org. In that case, we should show them a different end screen.

PR #632, opened 2022-08-25
Add sponsorless end screen

henriquelalves:
- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #609 

Adds a "SponsorlessEndScreen" Scene, that is shown if user is on the browser version of the app AND the domain site is "godotengine.org".

The screen itself is just an example, it may need modifications. Right now is very similar to the original screen, but without the sponsor-related text (it does invite users to contribute to the project's repository).

Nope.


NathanLovato:
From the docs, the way we expect this to work is users have a link and land on `https://gdquest.github.io/learn-gdscript/`

So we can't use the domain name. We need to use a parameter or something in the URL (E.g. `https://gdquest.github.io/learn-gdscript/?ref=godot-docs`) and save the fact the user started the course on docs.godotengine.org in their profile, so that even if they finish the course, say, 1 week later, they get the non-sponsored end screen.

Xananax:
In regards to Javascript;
the code is good, but to stay more consistent with the rest of the project:

1. use `OS.has_feature("JavaScript")` instead of `OS.has_feature("HTML5")`. In practice, the two feature flags will be equivalent 100% of the time, but `"Javascript"` is what is used elsewhere in the project, allowing to find all instances of JS usage quickly with [a global search](https://github.com/GDQuest/learn-gdscript/search?q=%22OS.has_feature%28%22JavaScript%22%29%22).
2. Other instances of JS usage (accessible through the search link above) include the necessary JS at the usage location, instead of externalizing it. If this function is to be used only in that instance, maybe there is no need for a standalone file? Doing a feature check when needed is good enough.
Alternatively, I'd suggest to add a method `is_godot_org_domain` in `JavascriptUtils` which returns a boolean in all cases, because [feature checking twice](https://github.com/henriquelalves/learn-gdscript/blob/fix/609-change-end-screen/ui/UICore.gd#L136) makes abstracting the function a bit redundant.
If `JavascriptUtils` is kept, we could consider moving some of the JS used elsewhere to it.

Note: From Godot 3.4 onwards, JS interfaces are supported natively, so you can do:

```gdscript
var window := JavaScript.get_interface("window")
window.location.href()
```
There's nothing wrong with using `eval`, but I just think it's neat.

henriquelalves:
Okaydokey, I'll refactor it a bit then!
Instead of getting the domain name, I'll look for `ref=godot-docs` in the URL path, is that ok?

henriquelalves:
@NathanLovato @Xananax done! I removed the "JavascriptUtils" class, since the code could be simplified further (and checking other places where it was checking for Javascript, it doesn't seem it made sense to decouple it to a different script)

PR #635, opened 2022-08-29
always use debug build

### LINKED ISSUE #618: Interpreter freezes upon trying to call non-existing member functions
**Describe the bug**
Calling a non-existing member function in any lesson freezes the whole app with an uncaught runtime error popping up in the (browser) console: `null function or function signature mismatch`.

**To Reproduce**
Steps to reproduce the bug:
1. Open any practice.
2. Define a new variable or use an existing one and call any non-existing member.
3. Click on `Run`.
4. The app freezes, triggering the runtime error.

**Expected behavior**
The non-existing member should be noted instead of the app freezing.

**Screenshots**
While this example is artificial, I've noticed it by making a typo using `pop_font()` rather than `pop_front()`.
gdquest-bug

 - Latest Google Chrome (stable)

Xananax:
fixes #618

Should be tested on staging first

### DIFF
--- a/run
+++ b/run
@@ -195,9 +195,14 @@ export_custom(){
   dir=$(dirname "$output")
   mkdir -p "$dir"
   prepare_versionfile
+  # THERE IS NO DIFFERENCE BETWEEN A DEBUG AND A RELEASE BUILD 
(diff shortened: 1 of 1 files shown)


NathanLovato:
Tested, working well!

Issue #636, opened 2022-08-30
Add examples unrolling loops in lessons and 17 and 18


NathanLovato:
Lesson 17 introduces while loops. It can be difficult to visualize what the code is doing, so it would be good to add a code example showing exactly all the instructions that run during the first loop example.

Same for lesson 18 and the first for loop example.

NathanLovato:
Done for 18, left to double check in 17, we need to ensure it's not confusing with while loops.

PR #641, opened 2022-09-01
Improvements to lesson 17 and 18

### LINKED ISSUE #589: Lesson 17 - Reason for moving one less than the board's size is unclear
The explanation that "we move one less than the board's width because the robot is already on the first cell" is unclear and makes it sound as though the robot's starting position matters (suggesting we would move two less than the board's width if the robot started on the second cell, which is not correct). If I'm understanding correctly, the actual reason we move one less than the board's width is because the equivalent robot position is one less than the board's width (or height) due to each being counted differently - A (0,0) board doesn't exist and a (1,1) board only has one square, but for the robot (0,0) means being on the first square. Which is in itself confusing, it would make far more sense if the cell variable being (1,1) meant the robot was on the first square but I don't know if it can be made to work that way.

Also, in the practice it would help if the description said that the robot's position is stored in the "cell" variable. This is shown if you scroll to the Documentation section, but it would still be easier to understand if instead of saying "The robot's current cell is Vector2(2, 0)." it said "The robot's current cell is Vector2(2, 0) and is stored in the cell variable."

### LINKED ISSUE #593: Lesson 18 - For Loop lack of clarity regarding the temporary variable
At the start of the lesson, an example is shown where a variable called "number" is created (the name of which I think has the potential to be confusing, but that's a different issue). Then, when showing the for loop, the temporary variable it uses is also called "number". Using the same name in both instances may cause confusion for some people, as they may misunderstand and think that the variables are the same (meaning that the "number" in the for loop is a variable that was previously declared in the code).

I would recommend changing this so that the examples showing the for loop use a different variable name than the first example that declares a variable. Also, for added clarity, in the part that says "In a for loop, the computer takes each value inside a list, stores it in a temporary variable, and executes the code in the loop once per value" it should be clarified in parentheses after "stores it in a temporary variable" what exactly that temporary variable is. If I'm understanding how this works correctly, it could maybe say something like "stores it in a temporary variable (which we name between the words "for" and "in" - note that this must follow normal variable naming rules, including the need for a unique name)".

### LINKED ISSUE #612: Lesson 18 references a `move_to_end()` function with a `while` loop which isn't the same
**Issue description:**

The lesson claims that this is the function that we used earlier:

img1

But in fact, it's actually this (in lesson 17):

img2

### LINKED ISSUE #615: Lesson 17 clarity
For those of us with thick skulls it would be more helpful for the cell Vector2 to be called something more descriptive such as, current_cell, or cell_position. The word "cell" in this instance doesn't let us know we should be thinking about the cell the ROBOT IS IN.

### LINKED ISSUE #626: Chapter 18: Why "number" in "for number in range()"
**Issue description:**
For me it wasn't clear enough why I use the identifier number. It didn't make a connection for me.
In the practice I use cell.y instead, cause I thought I want to move the robot Vector.y in rage of the board_size.
Here are screenshot what I did.

Maybe after I saw the solution and go through the lesson again and again. I kinda understand now. But it didn't really clicked. 

Screenshot 2022-08-19 225122

Screenshot 2022-08-19 224928

image

henriquelalves:
fix #626
fix #612
fix #593
fix #615
fix #589

Related issue (if applicable): #636 , #626, #612, #593, #615, #589 
There were a lot of small but good suggestions for lessons 17 and 18, this pull request tries to solve most of them. 

- Fix code highlight on lesson 17 on while-loop "step"
- Add some clarification on `cell` in lesson 17
  - I didn't go further with the suggested changes from #615 because that would imply changing the exercise and all "cell" references of Lesson 18 too; since it would be a more complex refactor (if taking into consideration changing `cell` to `current_cell` in the real code as well), I thought it would be best to leave this to a different PR.
- Add some clarification on starting counting cells at 0 in lesson 17
- Add some clarification on the loop code in lesson 18
- Fix some code illustration in lesson 18

This **does** add some repetition between lesson 18 and lesson 19 (it's mostly the same thing and examples on both lessons), but lesson 19 later explains arrays more thoroughly, while lesson 18 just use them on loops.

Also, I couldn't figure out a way to "unwrap" lesson 17 while-loops without giving room for misinterpretation. The problem is that while-loops branch out only when the condition is not true anymore, unlike the for-loops we used as examples that had a compile-time finite amount of loops. So unwrapping the while-loop as:

```
if value < 4:
  print(value)

if value < 4:
  print(value)

...
```
May give the impression that it works similar as the for-loop in the sense that `while value < 4` will always repeat the code for times max.

Nope.


course/lesson-18-for-loops/lesson.tres
```diff
@@ -122,9 +122,49 @@ title = ""
 type = 0
 text = "In the above example, for each item in the list [code][0, 1, 2][/code], Godot sets [code]number[/code] to the item, then executes the code in the [code]for[/code] loop.
 
-In this example, we're printing the value of [code]number[/code] as Godot moves through the loop.
+We'll get to explain arrays more throughly in the next lesson, but pay notice on the fact that [code]number[/code] is just a variable you are assigning values to using the [code]for[/code] loop, and you could name it anything you want. The above code has the exact same behaviour as the next example:"
```
NathanLovato, line 125: pay notice of the fact that -> redundant, maybe incorrect, but for conciseness, just `notice that` is good. Avoid extra words like "more" or "just"

NathanLovato: Other conciseness examples: exact same -> same

NathanLovato: > The above code has the exact same behaviour as the next example

Edited:

> This code behaves the same as the previous example

PR #642, opened 2022-09-01
Fix navigating multiple times

henriquelalves:

- For bug fixes and features:
    - [x] You tested the changes.

When clicking a button to navigate to a different screen multiple times quickly (sometthing users might do if the screen transition is slowish; I did this with a loading time of ~1sec), the next screen will appear repeatedly, causing a lag and bugging the breadcrumbs (only one instance of the next screen will appear in the breadcrumbs, not the whole stack of repeated screens). Lessons in particular will also appear with glitched text (probably something related to the bbcode formatting regex).

This fixes this behaviour. It creates a safe variable on _navigate_to() so it only navigates to the next screen after it finishes the transition the first time it was called.

Nope.

### DIFF
--- a/ui/UINavigator.gd
+++ b/ui/UINavigator.gd
@@ -24,7 +24,6 @@ var _matches := {}
 
 var _lesson_index := 0
 var _lesson_count: int = 0
-var _is_navigating_to: bool = false
 
 onready var _home_button := $Layout/Header/MarginContainer/HeaderContent/HomeButton as Button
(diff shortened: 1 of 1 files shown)


NathanLovato:
It's good to fix, but avoid adding extra state as much as possible. Having a new boolean not only adds extra code, but it opens up opportunities for more bugs and mistakes over using existing state.

Here, we can leverage the tween which is animating until the end of any transition. (with _tween.is_active())

There is another function for navigating back that wasn't covered, and while it was harder to trigger, clicking it fast while navigating would cause the next screen to get erased.

Issue #639, opened 2022-08-31
Make lesson 15: modulo an annex/bonus lesson


NathanLovato:
There are very few uses of modulo in Godot due to alternate functions being available. While the concept can be useful in code, it shouldn't be in the middle of the course, it's a bit out of place.

See issues #586 and #574

Issue #574, opened 2022-06-29
Lesson 15


DexLovesGames:
<Read the edit at the bottom, please.>
> Notice how the modulo can be larger than the number it affects. For example, 1 % 2 gives you 1. That's because 1 divided by 2 equals 0, and the remainder is 1.

> 1 divided by 2 equals 0

> 1 / 2 = 0

What? This doesn't make any sense. 1 divided by 2 equals 0.5, not 0, so how does the remainder end up being 1, shouldn't the remainder be 0.5 as well? This needs better explanation.

![image](https://user-images.githubusercontent.com/108374760/176326573-c2d9fea8-d63d-4d52-8cb5-75380cc8fe3b.png)

EDIT: Okay So I am dumb, and forgot how division without decimals works, but I still feel like better explanation of this would be helpful. I realize now that 3 / 2 = 1 with a remainder of 1, and the remainder is all that 3%2 is looking for, and if I apply the same logic to 1/2 then 2 goes into 1 0 times, with 1 being the integer left over as remainder. Basic math, and here I'm trying to look at it from college level algebra lol.

PR #654, opened 2022-09-08
Add Debugger-like visualization to variables

henriquelalves:
This Pull Request adds:
- A tooltip for variables showing their value while the code is running.
- A "debugger" visualization in which users can check the script variables values while the example runs.

### DIFF
--- a/ui/components/RunnableCodeExample.gd
+++ b/ui/components/RunnableCodeExample.gd
@@ -262,53 +262,60 @@ func _set_scene_instance(new_scene_instance: CanvasItem) -> void:
 	if not _run_button.visible:
 		printerr(ERROR_NO_RUN_FUNCTION % [_scene_instance.filename])
 
+	# Setting up our fake debugger when it's there to allow executing the code line-by-line
+	var debugger: RunnableCodeExampleDebugger = null
 	for node in get_parent().get_children():
 		if node is RunnableCodeExampleDebugger:
-			var debugger: RunnableCodeExampleDebugger = node
-			debugger.setup(self, _scene_instance)
+			debugger = node
+			break
 
-			if _scene_instance.has_signal("code_updated"):
-				_scene_instance.connect("code_updated", self, "emit_signal", ["code_updated"])
-			_gdscript_text_edit.update()
-			yield(get_tree().create_timer(0.1), "timeout")
-			if _gdscript_text_edit.visible:
-				var monitored_variables : Array = debugger.monitored_variables
-				var offset := Vector2.ZERO
-				offset.x = _gdscript_text_edit.rect_position.x
+	if not debugger:
+		return
 
-				for variable_name in monitored_variables:
-					var last_line := 0
-					var last_column := -1 # Search offset to not repeat same result
+	debugger.setup(self, _scene_instance)
+	if _scene_instance.has_signal("code_updated"):
+		_scene_instance.connect("code_updated", self, "emit_signal", ["code_updated"])
 
-					while last_line >= 0:
-						var result := _gdscript_text_edit.search(variable_name, 0, last_line, last_column + 1)
-						
-						var is_result_in_line_before := false
-						var is_result_in_column_before := false
-						
-						if result.size() != 0:
-							is_result_in_line_before = result[TextEdit.SEARCH_RESULT_LINE] < last_line
-							is_result_in_column_before = (
-							result[TextEdit.SEARCH_RESULT_COLUMN] < last_column and
-							result[TextEdit.SEARCH_RESULT_LINE] <= last_line)
-						
-						if result.size() == 0:
-							last_line = -1
-						elif is_result_in_line_before or is_result_in_column_before:
-							last_line = -1
-						else:
-							last_line = result[TextEdit.SEARCH_RESULT_LINE]
-							last_column = result[TextEdit.SEARCH_RESULT_COLUMN]
+	if not _gdscript_text_edit.visible:
+		return
 
-							var rect = _gdscript_text_edit.get_rect_at_line_column(last_line, last_column)
-							rect.position += offset
-							rect.size.x = (rect.size.x * variable_name.length()) + 4
+	# Create widgets that underline a variable and display a variable's value
+	# when hovering with the mouse.
+	var monitored_variables := debugger.monitored_variables
+	var offset := Vector2(_gdscript_text_edit.rect_position.x, 0.0)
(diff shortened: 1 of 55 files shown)


Xananax:
It's very cool! Impressive work.

Do I understand correctly that the example implementation is in "Chapter 18 - introduction to For Loops", scene `res://course/lesson-18-for-loops/visuals/PrintForLoop.tscn`?

If yes, it seems the hover popups don't show in situ (they show when I run the scene with F6).

Apart from that, I have some comments regarding the code.

Currently, if I understand properly, for the debugger to work, you need two dependent, but independently entered pieces of information:

- export a `monitored_variables` array
- add a `RunnableCodeExampleDebugger` node

If that is correct, I would suggest making it so it's not possible to have one without the other, to minimize both necessary documentation, and possible mistakes. For example, we could:

1. Add the `RunnableCodeExampleDebugger` node at runtime, if `monitored_variables` is found
2. Add the `monitored_variables` export to `RunnableCodeExampleDebugger` and read it from there

Of these 2, I think the second is preferable, because of its greater clarity and discoverability. It also allows to reshape the debugger's size or theme if necessary.

henriquelalves:
Completely agree with you there. I don't like the first refactor option because to add the DebuggerScene, I have to replace the root of the example scene to a VBoxContainer, and making this change may be really cumbersome.

I refactored using your second idea. The only problem is that we can only highlight variables when the DebuggerScene is present, but I don't know if we ever want to highlight variables and don't add a debugger scene anyway.

I'm also adding the Debugger scene on lesson 18 examples to test it, but I can remove it before merging. Weird thing is that I'm able to see it on lesson 18 though, can you confirm the error of the examples not having the debugger scene on the lesson 18?

Xananax:
Regarding the pop-up: I do see the _debugger_, but I couldn't see the _pop-up number_ on hover.

Trying again, it appears the issue is if I press `run()` before pressing `step()`. The highlights and numbers then stop appearing. If I press only `step`, then all is good.

Apart from that, I got nothing, it's looking good. I didn't do a full code review, once the bug is fixed, please add me as a reviewer or ping me on discord, I'll go through the code with more scrutiny.

On another hand: Should implementation sites be part of this PR, or is lesson 18 the only implementation site?

Once the bug is fixed, I suggest using the debugger, in this same PR, in all the lessons where it can/should be used, as to uncover more bugs and use cases before merging.

If you decide to do that, then we can review the code afterward (it might change).

henriquelalves:
Just updated all relevant lessons! My priority was to change lessons with more than one or two lines of code, in which a variable changes value. I also fixed one or two issues along the way, so now all lessons should be working.

henriquelalves:
Tagging related content issues this PR may have affected/solved.

- Lesson 8
  - #659 
- Lesson 15
  - #639 
  - #574 
- Lesson 17
  - #636

Xananax (commented):
Excellent! This is really neat.

Apart from a few changes to the types, I didn't have anything to review.

A few comments/questions

## 1 - Bug Report

Currently, pressing "reset" doesn't seem to reset the value, at least in that exercise

debugger

I did not check the others

## 2 - Documentation

I can guess the process to add the debugger and the arrow highlighter, but I'd like a formal description of the process which we can give in general to people who want to use it.

A few bullet points is enough, it doesn't need to be detailed or in-depth, it's for internal usage

## 3 - Automation

For the purposes of the current project, I think the scope of this is very good, but it seems to me like the process is involved and has several points where errors can easily happen (I believe that's the reason for the bug described above). For a potential future update, do you envision any way to automate completely or partially the process, to reduce potential errors and simplify workflow?

Alternatively or additionally, is there any way to error check better that everything is set up correctly?

henriquelalves:
# 1 - Bug Report
Good catch! Didn't notice there was a bug with reset(). I had to fix it on all examples by overriding the "reset()" function and setting the variable to their initial value - it's one more thing to remember when adapting a code example, but I don't think there is any other way to do this without making it way less clear how the debugger interacts with the code.

# 2 - Documentation
Just created this small gist with some step-by-step instructions, I can update it if there is any questions on how to change an example: https://gist.github.com/henriquelalves/365f46893e99d8b6cacfeb37d4901399

# 3 - Automation
The steps to adapt an example to add a Debugger is somewhat straightforward, but I think we could automate it with an addon to make it simpler (changing the root of the example to a VBox and addding a Debugger scene). I think what is error-prone is adapting the examples to run in step-by-step mode, but I don't know if there is an easy way to automate it without making the whole process much less clear on how it works.

NathanLovato:
When testing or pushing a PR, be sure to test every single change in the context the user will see it. I.e. if you tested individual examples by pressing F6, don't do that. In L15 for instance, the traffic light demo is cut off:

image

In lesson 13 there are these visual artifacts

image

I spotted a couple of little things like these re-reading the full lessons, this still lacks some cleanup

NathanLovato:
Getting a weird issue: the variable highlight breaks in Godot 3.5, but not always, and displays like this.

image

henriquelalves:
Just added an "UNINTIALIZED_VARIABLE_VALUE" constant  on the Debugger, so I'm hiding the variable if it's the unintialized string. I also updated the `reset()` function to always be called once before running (we were already doing that but implicitly, since we were running "clear_messages()" when the function started).

ui/components/RunnableCodeExample.gd
```diff
@@ -313,11 +326,12 @@ func _set_scene_instance(new_scene_instance: CanvasItem) -> void:
 				rect.position += offset
 				rect.size.x = (rect.size.x * variable_name.length()) + 4
 
-				var monitored_variable: CodeExampleVariableUnderline = CodeExampleVariableUnderlineScene.instance()
```
NathanLovato: I invite you to use the formatter gdformat. Without space is the official gdscript style guide

ui/components/RunnableCodeExample.gd
```diff
@@ -267,26 +270,36 @@ func _set_scene_instance(new_scene_instance: CanvasItem) -> void:
...
+	for monitored_variable in _monitored_variable_highlights:
+		monitored_variable.queue_free()
+	_monitored_variable_highlights.clear()
 
 	if not _gdscript_text_edit.visible:
 		return
 
 	# Create widgets that underline a variable and display a variable's value
 	# when hovering with the mouse.
-	var monitored_variables := debugger.monitored_variables
```
NathanLovato: You're getting an error with type inference here? it's working fine on my end

PR #655, opened 2022-09-08
Improve Lesson 22 cell screen-position explanation

henriquelalves:
Adds text explanation to Lesson 22 to improve clarity on the conversion of Cell coordinate to screen position.

### DIFF
--- a/course/lesson-22-functions-return-values/lesson.tres
+++ b/course/lesson-22-functions-return-values/lesson.tres
@@ -154,7 +154,7 @@ In previous lessons, we had characters walking on grids.
 
 And for those practices, you were working directly with cell coordinates.
 
-Well, cell coordinates don't correspond to positions on the screen. Each cell have a size in pixels of 120x120, so to get the center of any cell, we need to convert it's cell position to a screen position."
+Well, cell coordinates don't correspond to positions on the screen. To find the center of any cell on the screen, we need to convert the cell's coordinates to a position on the screen, in pixels."
 visual_element_path = ""
 reverse_blocks = false
 has_separator = false
--- a/course/lesson-22-functions-return-values/lesson.tres
+++ b/course/lesson-22-functions-return-values/lesson.tres
@@ -174,7 +174,12 @@ script = ExtResource( 4 )
 content_id = "res://course/lesson-22-functions-return-values/content-BWNbDGt6.tres"
 title = ""
 type = 0
-text = "To do so, we use a function. It does two things: first, it multiplies the cell coordinate by the cell size which gets us it's top-left corner posititon in pixels. Then it adds half of it's size to get the center of the cell, and finally returns the result."
+text = "To do so, we use a function. The function does two things:
+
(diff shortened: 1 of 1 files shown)


course/lesson-22-functions-return-values/lesson.tres
```diff
@@ -154,7 +154,7 @@ In previous lessons, we had characters walking on grids.
 
 And for those practices, you were working directly with cell coordinates.
 
-Well, cell coordinates don't correspond to positions on the screen, so we need to convert cell positions to screen positions."
+Well, cell coordinates don't correspond to positions on the screen. Each cell have a size in pixels of 120x120, so to get the center of any cell, we need to convert it's cell position to a screen position."
```
NathanLovato, line 157:
```suggestion
Well, cell coordinates don't correspond to positions on the screen. To find the center of any cell on the screen, we need to convert the cell's coordinates to a position on the screen, in pixels."
```

NathanLovato: Fixing two grammar errors:

- `each` is singular -> `each cell has`
- `it's position` -> `its position` because it's a possessive pronoun and not the verb to be we're using here.

I recommend running your edits through a grammar checker. You could use one in emacs, though options are pretty limited. A simple solution is to copy paste your writings into [hemingway](https://hemingwayapp.com/) or google docs.

course/lesson-22-functions-return-values/lesson.tres
```diff
@@ -174,7 +174,7 @@ script = ExtResource( 4 )
 content_id = "res://course/lesson-22-functions-return-values/content-BWNbDGt6.tres"
 title = ""
 type = 0
-text = "To do so, we use a function. It multiplies the cell coordinate by the cell size, adds half the cell size to the product, and returns the result."
+text = "To do so, we use a function. It does two things: first, it multiplies the cell coordinate by the cell size which gets us it's top-left corner posititon in pixels. Then it adds half of it's size to get the center of the cell, and finally returns the result."
```
NathanLovato, line 177:
```suggestion
text = "To do so, we use a function. The function does two things:

1. First, it multiplies the cell coordinates by the cell size, which gives us the position of the cell's top-left corner on the screen, in pixels.
2. Then, we add half of the cell size to get the center of the cell.

The function returns the result, allowing us to store it in a variable."
```

NathanLovato: This needed more thorough changes. When we talk about doing X steps, when possible it's good to use numbered lists, they help with reading. And that even if we didn't have them in the original lessons, making improvements we forgot is always good.

course/lesson-22-functions-return-values/lesson.tres
henriquelalves: That's a good idea, I'll start double checking my writing with Google Docs.

course/lesson-22-functions-return-values/lesson.tres
henriquelalves: This is much better, thanks!

PR #660, opened 2022-09-18
Write a script to split translation files with multiline translations into single-line ones

### LINKED ISSUE #556: Split multi-line translation strings into single line translations
The translation strings currently take strings from the content blocks as-is but it makes it more difficult to make and update translations. It gives extra work to translators (multiplied by content updates and languages).

We need to split blocks into single-line translations instead.

Tasks:

- [x] Split the lines from the content blocks when extracting strings
- [ ] Split content block lines when populating lessons, practices, etc., and call the tr() function on each line.
- [ ] Use a script to merge all existing translations into the new translation file structure

NathanLovato:
Close #556

### DIFF
--- a/i18n/extract.py
+++ b/i18n/extract.py
@@ -99,7 +99,7 @@ def extract_babel_and_write(
 ) -> None:
 
     print("  Starting extraction...")
-    cat = Catalog(
+    catalog = Catalog(
         project=PROJECT,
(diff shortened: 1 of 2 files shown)


NathanLovato:
The script to split PO files is done, now we need to generate POT files accordingly when extracting translations from Godot

PR #664, opened 2022-09-19
Improve clarity on lesson 22 content about documented functions

### LINKED ISSUE #581: Lesson 22 discrepancy
At the end of the lesson, before practice it says this:
image

This implies that we should check documentation first in this practice,
but then, in the practice itself, there isn't any documentation for any functions at all.
image

henriquelalves:
Fix #581 

Change sentence regarding documentation in practices to improve clarity.

### DIFF
--- a/course/lesson-22-functions-return-values/lesson.tres
+++ b/course/lesson-22-functions-return-values/lesson.tres
@@ -46,7 +46,7 @@ script = ExtResource( 4 )
 content_id = "res://course/lesson-22-functions-return-values/content-ZPxY8VUD.tres"
 title = ""
 type = 0
-text = "Imagine you have a game where you track the player's health as a percentage, a decimal number going from [code]0.0[/code] to [code]100.0[/code]. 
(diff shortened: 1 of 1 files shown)


course/lesson-22-functions-return-values/lesson.tres
```diff
@@ -219,7 +219,7 @@ script = ExtResource( 4 )
 content_id = "res://course/lesson-22-functions-return-values/content-2rtQEn91.tres"
 title = ""
 type = 0
-text = "Some functions return values, and some do not. You can learn which functions return a result using the documentation panel in the practice screen.
+text = "Some functions return values, and some do not. During practices, you can learn which functions return a value using the documentation panel when provided.
```
NathanLovato, line 222:
```suggestion
text = "Some functions return values, and some do not. During practices, you can learn which functions return a value using the documentation panel. It will display if the practice requires using specific functions or variables.
```

PR #665, opened 2022-09-19
Add IN_EXPECTED_AFTER_IDENTIFIER error

### LINKED ISSUE #625: Introduction to For Loops - Missing Error Describtion
**Describe the bug**
Missing Error describtion for  **"in" expected after identifier**.

**To Reproduce**
1. Go to Capter 18
2. Click on the first practice 
3. Type in what I done, showed in the screenshot

**Expected behavior**
Explaination for the ERROR

**Screenshots**
image

**Information about your device:**
 - Operating System: Windows
 - Browser: No Browser - In App

henriquelalves:
Fix #625 

Adding an error when the user doesn't write a valid identifier in a `for` loop. Had to check when Godot throws this error, but it basically happens when the parser finishes interpreting an identifier after a `for` and it's expecting an `in`; so the user may have used punctuations or spaces in the identifier, or forgot to add an `in` to the `for` loop.

### DIFF
--- a/script_checking/error_database.csv
+++ b/script_checking/error_database.csv
@@ -30,9 +30,14 @@ INVALID_OPERATOR_USAGE,,
 MISPLACED_STATIC_CALL,,
 IN_EXPECTED_AFTER_IDENTIFIER,"You get this error when the name between the [code]for[/code] and [code]in[/code] is not a valid variable name, or you are missing the [code]in[/code] keyword.
 
-In a [code]for[/code] loop, the [code]in[/code] keyword only accepts a valid variable name to iterate. It will create a new variable with the given name, and it will assign each element of the array to it.","To fix this error, you need to ensure that the name between [code]for[/code] and [code]in[/code] is a valid variable name, with no punctuations or spaces in-between the word.
+In a [code]for[/code] loop, the [code]in[/code] keyword only accepts a valid temporary variable name to assign values in each loop iteration. The loop creates a new variable with the desired name and assigns each element of the array to it.","To fix this error, you need to ensure that the name between the [code]for[/code] and [code]in[/code] keywords is a valid variable name with no punctuation or spaces.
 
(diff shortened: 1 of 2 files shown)


script_checking/error_database.csv
```diff
@@ -26,6 +26,13 @@ UNSAFE_CAST,,
 UNSAFE_CALL_ARGUMENT,,
 DEPRECATED_KEYWORD,,
 STANDALONE_TERNARY,,
+INVALID_OPERATOR_USAGE,,
+MISPLACED_STATIC_CALL,,
+IN_EXPECTED_AFTER_IDENTIFIER,"You get this error when the name between the [code]for[/code] and [code]in[/code] is not a valid variable name, or you are missing the [code]in[/code] keyword.
+
+In a [code]for[/code] loop, the [code]in[/code] keyword only accepts a valid variable name to iterate. It will create a new variable with the given name, and it will assign each element of the array to it.","To fix this error, you need to ensure that the name between [code]for[/code] and [code]in[/code] is a valid variable name, with no punctuations or spaces in-between the word.
+
+You may want to assign something to the property of the variable (for example, in a [code]Vector2[/code] array). In this case, first assign a variable name on the loop, and then change the property inside the loop."
```
NathanLovato, line 35:
```suggestion
IN_EXPECTED_AFTER_IDENTIFIER,"You get this error when the name between the [code]for[/code] and [code]in[/code] is not a valid variable name, or you are missing the [code]in[/code] keyword.

In a [code]for[/code] loop, the [code]in[/code] keyword only accepts a valid temporary variable name to assign values in each loop iteration. The loop creates a new variable with the desired name and assigns each element of the array to it.","To fix this error, you need to ensure that the name between the [code]for[/code] and [code]in[/code] keywords is a valid variable name with no punctuation or spaces.

For example, this code is invalid: [code]for cell_position.x in cell_positions_array:[/code] because [code]cell_position.x[/code] isn't a valid variable name.

To access the [code]x[/code] sub-component of the variable, you need to do that inside of the loop's body:

[code]for cell_position in cell_positions_array:
    cell_position.x += 1.0[/code]"
```

Issue #685, opened 2022-10-02
Glossary does not work after switching languages


Wolfy7:
**Describe the bug**
After switchig the language to Spanish and the back to English the glossary does not work anymore.

**To Reproduce**
Steps to reproduce the bug:
1. Switch language to Spanish
2. Apply
3. Switch back to English

**Expected behavior**
Even after language swap all words from the glossary should be highlighted as such and clickable.

**Screenshots**
At the beginning, glossary works as expected:
![image](https://user-images.githubusercontent.com/16044116/193474841-8b9480d1-57a4-47e7-9a18-fb969b6b7f87.png)

Switching language to Spanish:
![image](https://user-images.githubusercontent.com/16044116/193474854-dbffcfbf-6b0a-4033-8e9a-7e50f6e3b95a.png)

After switching back to English, glossary does not work anymore:
![image](https://user-images.githubusercontent.com/16044116/193474864-fb1a6884-6ffb-47b9-9f81-0cbe84a733ff.png)

**Information about your device (please complete the following information):**
 - Operating System: Windows 10 (21H2)
 - Godot-Version: 3.4.5

PR #687, opened 2022-10-02
feat: make glossary entries work in other languages

Wolfy7:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #685

Emit the "translation_changed" signal from the TranslationManager when the language is set.
Listen to this signal in UILesson.gd and trigger the new defined function `_underline_glossary_entries` for Underline glossary entries.

### DIFF
--- a/autoload/TranslationManager.gd
+++ b/autoload/TranslationManager.gd
@@ -52,6 +52,7 @@ func set_language(language_code: String) -> void:
 		TranslationServer.set_locale(current_language)
 		current_profile.language = current_language
 		current_profile.save()
+		emit_signal("translation_changed")
 		return
 	
 	# Load order shouldn't be important, so we'll just load everything from the folder.
@@ -106,6 +107,7 @@ func set_language(language_code: String) -> void:
 	TranslationServer.set_locale(current_language)
 	current_profile.language = current_language
 	current_profile.save()
+	emit_signal("translation_changed")
 
 
 func _reset_language() -> void:
--- a/resources/Glossary.gd
+++ b/resources/Glossary.gd
@@ -68,6 +68,6 @@ class Entry:
 	var explanation: String
 
 	func _init(csv_line: Array) -> void:
-		term = tr(csv_line[0].capitalize())
+		term = tr(csv_line[0]).capitalize()
 		plural_form = tr(csv_line[1])
 		explanation = tr(csv_line[2])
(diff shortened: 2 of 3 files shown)


Wolfy7:
If changes are okay, please label it for **Hacktoberfest**
Thank you 😊

NathanLovato:
Thanks for the help! This currently doesn't work because it tries to underline the English keywords in other languages. For example, in Spanish, "function" becomes "función" so we would need to target this keyword. To achieve that,  you'll need to update `_glossary.replace_matching_terms()` (or generally the glossary script) to compile and use regexes in the target language.

The way it works right now is we make a big regex with all the glossary terms to find and use a regex replace to find and underline them. But it's only done in English.

Wolfy7:
Yes, you right. 
This will just fix that if you switch the language from english to another language and back again to english, that then the word from the glossary are underlined again.

To handle this for other languages like Spanish, there are more work to do. As you already wrote.

I will think about it how to archive this, if you have already something in mind i would appreciate if you shared it with me 😊

Should this done with an addtional PR or in this one?

Thank you for your help.

NathanLovato:
Thanks. I'd prefer this to be done in this PR because otherwise, fixing this bug the way it's currently done will introduce a separate bug for people using other languages. We might as well handle both cases at once.

Wolfy7:
If we create/read the glossary after TranslationManager is ready and also recreate it after a languange switch. 
It should work without changing `_glossary.replace_matching_terms()`

But not completely sure if these changes are wanted.

NathanLovato:
Rebuilding the glossary on a language change and changing the highlighted keywords at that time sounds good to me, it's a good approach!

Wolfy7:
I did these changes in my last commit https://github.com/GDQuest/learn-gdscript/pull/687/commits/959c6701637fdab074be667a294cd218ced2d0f3

In my test it works as expected:
image

image

NathanLovato:
Great work André. The issue was small, you were trying to translate the capitalized term in the glossary entries but then that would not match translation strings.

I just moved the capitalize() call after the tr() call.

image

Wolfy7:
Thank you Nathan.

Just want to use the change, writing with you. To say **Thank you!** of course to you and the whole GDQuest-Team.
I follow your YouTube-Channel since years, and every video is a pleasure to watch.

But also to all the work you  (GDQuest) put into the OpenSource-Community, the GameDevelopment community.
I'm pleasured to work on your amazing products.

So, yeah im not so good with words. But wanted to let you know. An just once again **Thank you.**.

NathanLovato:
Thanks :)

Issue #497, opened 2022-04-20
Add CJK font


NathanLovato:
People are translating the app to Chinese and Japanese. We should include fonts for those languages in the app for testing purposes.

EDIT: PR https://github.com/GDQuest/learn-gdscript/pull/688 attempted to add those. However, these fonts are heavy and we'd like not to make the initial download larger for every user., but rather do it on demand.

Fonts should be downloaded at runtime and cached when users switch to Chinese or another CJK language. This means they also need to be swapped at runtime, using Godot's theming API.

Wolfy7:
Hi Nathan.
Just for understanding, purpose of this issue is "just" to add some addtional fonts for support this languages. Right?
So for example put a font from here: https://fontmeme.com/fonts/cjk-fonts-collection/ under the folder "ui/theme/fonts"

NathanLovato:
Yes, this also involves updating many font resources with the appropriate font files/fallbacks. See `res://ui/theme/fonts`. Ideally I'd like a CJK font that's not only good-looking, clearly readable, but also fairly lightweight if possible. A great candidate is Noto Sans: https://fonts.google.com/noto

NathanLovato:
And as far as languages are concerned, we will need at least Japanese, simplified Chinese, perhaps Korean. We currently can't support arabic, at least iirc right-to-left writing isn't supported in Godot 3, so it's something we'll have to add when porting to Godot 4.

Wolfy7:
Hi. Thanks for make it clearer.

I had then just some addtional quqstions:

- Yes, the Noto Sans looks good therfore (i also had seen this). There is also Noto Serif. 
  - So these Fonts could be added:
    - Korean: https://fonts.google.com/noto/specimen/Noto+Sans+KR?noto.lang=ko_Kore&noto.continent=Asia&noto.script=Kore
    - Japanese: https://fonts.google.com/noto/specimen/Noto+Sans+JP?noto.query=Japanese&noto.lang=ja_Jpan&noto.script=Jpan
    - For Chinese there are two, one for  Simplified Chinese and one for Traditional Chinese
- Should there are also new resources like `font_documentation_bold.tres` just then for the corresponding language for exaample `font_documentation_korean.tres`
- If i undestand the Documentation right:
> DynamicFont also supports defining one or more fallback fonts, which will be used when displaying a character not supported by the main font.
  - We should add then these new fonts as "Fallback" in all font resources 

[...]

PR #688, opened 2022-10-03
improvement: Add Chinese font

Wolfy7:

Related issue (if applicable): #497

Add some Traditional Chinese text to an Label with for example `font_documentation_bold.tres` as font.

image

**What is the new behavior?**

image

### DIFF
--- a/ui/theme/fonts/font_button.tres
+++ b/ui/theme/fonts/font_button.tres
@@ -1,9 +1,15 @@
 [gd_resource type="DynamicFont" load_steps=2 format=2]
 
 [ext_resource path="res://ui/theme/fonts/Montserrat-Bold.ttf" type="DynamicFontData" id=1]
+[ext_resource path="res://ui/theme/fonts/NotoSerifJP-Bold.otf" type="DynamicFontData" id=2]
+[ext_resource path="res://ui/theme/fonts/NotoSerifTC-Bold.otf" type="DynamicFontData" id=3]
+[ext_resource path="res://ui/theme/fonts/NotoSerifKR-Bold.otf" type="DynamicFontData" id=4]
 
 [resource]
 size = 24
 use_mipmaps = true
 use_filter = true
 font_data = ExtResource( 1 )
+fallback/0 = ExtResource( 3 )
+fallback/1 = ExtResource( 2 )
+fallback/2 = ExtResource( 4 )
--- a/ui/theme/fonts/font_button.tres
+++ b/ui/theme/fonts/font_button.tres
@@ -1,9 +1,7 @@
-[gd_resource type="DynamicFont" load_steps=2 format=2]
+[gd_resource type="DynamicFont" load_steps=3 format=2]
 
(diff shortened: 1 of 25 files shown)


NathanLovato:
Thanks much! The PR looks good, I just need to check something with my teammates.

@Xananax @Razoric480 do you know if we have a way to separate these fonts and download them only when needed (when changing the language to Chinese or Japanese, for example) in HTML5 builds? They're pretty big, 5-7mb a piece, and it's gonna get worse as we add languages (Cyrillic, Arabic, ...), I wouldn't want the initial download to be drastically slowed down ideally.

Xananax:
I will look into it, in theory we can add pcks and load them at will like addons, but I've never done it for web

Xananax:
Any idea how to author packages?

There's a recommended method: https://docs.godotengine.org/en/stable/tutorials/export/exporting_pcks.html

But I'm not sure how to extract the changed files as a package from the main project on authoring. Should the files be moved into a different project, then the package built? If there's no provision for this in Godot, should we do this ourselves?

I'll start by working on the loading part and verify it can work properly, we can then think about the authoring part.

If we agree, then I'll open a different issue to track progress.

NathanLovato:
And yeah sounds good to figure out authoring the pck after ensuring that loading in the browser works

Razoric480:
The ideal would be Godot supporting being able to white list directories, but so long as you're careful about what files are referenced, you can set the export mode of a given preset to selected resources and dependencies and only highlight the ones you want (and dis-highlight the ones you don't want from your main PCK export)

You can use download to `res://` using `HTTPRequest`, apparently:
```gdscript
func _ready():
    var client = HTTPRequest.new()
    client.connect('request_completed', self, '_request_completed')
    add_child(client)
    client.download_file = "res://name of your pck file.pck"
    client.request("http://link to your pck file")
    
    func _request_completed(result, response_code, headers, body):
    	if response_code == 200:
    		ProjectSettings.load_resource_pack("res://name of your pck file.pck")
    		get_tree().change_scene("res://name of your scene inside .pck file")
```
though I haven't tested it.

Wolfy7:
So are these changes okay and the seperation of the fonts will be handled in #691 ?

NathanLovato:
No, after spending some hours doing research on this it appears this approach will not work. We're going to have to handle everything in code, and the process is going to be quite a bit different. Sorry for having you work on this, it's unfortunately the kind of thing you don't know until you see the problems arise.

Wolfy7:
No, problem.
Do you already have an idea how you will handle it in code? If there is anything i can help with, let me now.

NathanLovato:
Looking back at this, I think we could just include the Chinese font, as we have complete or nearly complete Chinese translations. It'll increase the weight of the app a bit but it's probably better than not having Chinese support. We can forget about the dynamic download part.

We could add other fonts (Japanese and Korean) once we have complete translations for Japanese and Korean.

If you'd like I can take care of finalizing this for you. Just let me know.

Wolfy7:
@NathanLovato Sure, if you want., you can finilaize this.

Issue #623, opened 2022-08-14
light sensitivity and contrast


BloodyHazardous:
Being able to change the colour of the text would benefit me personally, recently reading has become a pain, due to increase light sensitivity, reading white writing on a dark background causes eyes issues, irritation. Its the contrast between light and dark. And due to this being a program and not a webpage there's not much i can do,

NathanLovato:
There's now an option for the main text thanks to #689. We're missing the code's colors to complete the feature. But this should already be easier on the eyes for most of the app's text.

russeg:
Not sure why this is the feature, since what should be implemented is Light/Dark theme switch.

Light mode should be the default, and dark an option since dark mode is actually an [accessibility problem](https://medium.com/@h_locke/why-dark-mode-causes-more-accessibility-issues-than-it-solves-d2f8359bb46a).

PR #689, opened 2022-10-03
feat: add option to change text color

Wolfy7:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #623

Adds new Setting **Text Color** to the SettingsPopup to change the color of all (`Label ` and  `RichTextLabel` ) text in the app.
Color setting is also stored to the `Profile`.

## New feature or change ##

image

image

image
image

**Other information**
Maybe its not the best idea to change `Label ` and  `RichTextLabel` or probably there are Control Nodes who should change color.

### DIFF
--- a/autoload/ThemeManager.gd
+++ b/autoload/ThemeManager.gd
@@ -3,6 +3,8 @@ extends Node
 const THEME_ROOT := "res://ui/theme/"
 const THEME_FONTS_ROOT := "res://ui/theme/fonts/"
 
+onready var _theme = preload("res://ui/theme/gdscript_app_theme.tres")
+
 var _font_size_defaults := {}
 
 
@@ -11,6 +13,7 @@ func _ready() -> void:
 	
 	var current_profile := UserProfiles.get_profile()
 	scale_all_font_sizes(current_profile.font_size_scale, false)
+	change_font_color(current_profile.font_color, false)
 
 
 func _cache_font_size_defaults() -> void:
@@ -57,3 +60,12 @@ func scale_all_font_sizes(size_scale: int, and_save: bool = true) -> void:
 		current_profile.font_size_scale = size_scale
 		current_profile.save()
 		Events.emit_signal("font_size_scale_changed", size_scale)
+
+func change_font_color(color: Color, and_save: bool = true) -> void:
+	_theme.set_color("font_color", "Label", color)
+	_theme.set_color("default_color", "RichTextLabel", color)
+	
+	if and_save:
+		var current_profile := UserProfiles.get_profile()
+		current_profile.font_color = color
+		current_profile.save()
--- a/autoload/ThemeManager.gd
+++ b/autoload/ThemeManager.gd
@@ -3,6 +3,9 @@ extends Node
 const THEME_ROOT := "res://ui/theme/"
 const THEME_FONTS_ROOT := "res://ui/theme/fonts/"
 
+const DEFAULT_COLOR := Color.white
+const LOWER_CONTRAST_COLOR := Color.darkgray
+
 onready var _theme = preload("res://ui/theme/gdscript_app_theme.tres")
 
 var _font_size_defaults := {}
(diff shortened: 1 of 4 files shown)


NathanLovato:
Thanks for your contribution!

Instead of giving a color picker, I would rather provide a text contrast slider or toggle button and map it to carefully hand-picked colors. This is to ensure that the user cannot, for example, pick a color that makes the app unusable. But also, they shouldn't be able to pick red or the purple color used in the background (which would make the text invisible).

On top of that, highlighted keywords in the text use different colors than the main text. They may need adjustments too, and having predefined colors depending on the desired contrast level would allow us to provide a cohesive color palette.

Writing this, I think that a "Lower contrast" toggle button would be a good option to try. Having a single low-contrast option that changes only the text color for now (later it could also tweak other colors).

Note that it needs to affect the lesson text, but also the TextEdit palette (code editor in the practice UI).

Wolfy7:
Yes, you right. 
During the implementation i also thought that this will be error prone. 

I also thought on something like diffrent color themes that could be chosse or something, but probably this is then too much to maintain.

A "Lower contrast" toggle button option sounds like a good approach, at least for the first step.
Do you have some "lower contrast" color already in mind? And probably you now which Control Notes (`Label `, `RichTextLabel`) should affected?

Thansk for helping.

NathanLovato:
> I also thought on something like diffrent color themes that could be chosse or something, but probably this is then too much to maintain.

Yes, that would be a neat feature, but it's extra work and maintenance for something low-priority right now.

>  Do you have some "lower contrast" color already in mind? And probably you now which Control Notes (Label , RichTextLabel) should affected?

Would you be comfortable writing the code and implementing the feature with placeholder colors? We could then jump in and tweak the color palette on top of your pull request.

Regarding the nodes, this will first need to affect Label and RichTextLabel.

Then there are various TextEdit nodes. Those are trickier. To understand why to see `CodeEditorEnhancer.gd`: The text edit class doesn't have a great API. Whenever we register a new keyword, we need to hard code the color of this keyword instead of using the theme. Also, it has to be done separately on every text edit node, and this can't be undone: to change the colors, you must delete the text edits and instantiate new ones. So to add support for changing the contrast of the highlighted keywords, I think that we will need more refactoring.

So in this is fine by you, I would recommend focusing on just the label and rich text label nodes to start with. And just the normal text color. Once this is done, we can consider the text edit nodes and keyword highlighting in lessons.

Wolfy7:
Yes, i could write the code with placeholders for the color(s).

But jsut to understand everything correctly, which approach we want now to start with:

- A "CheckButton" to just toogle between "normal" and "Lower contrast"?
  - So just one placeholder color is needed.
- Something like a slider or table to to select the "Lower contrast" color from some color suggestions

At the moment i already handle `Label `and `RichTextLabel`, so i then only have to replace the ColorPicker by the chosen approach.

After that i could check `CodeEditorEnhancer.gd` to get an overview how maybe TextEdit nodes could be handeld.

NathanLovato:
> A "CheckButton" to just toogle between "normal" and "Lower contrast"?

Yes! I think it's best to keep things as simple as possible by default and only add more options as people really need them.

NathanLovato:
> After that i could check CodeEditorEnhancer.gd to get an overview how maybe TextEdit nodes could be handled.

For this we will need to delete the existing text edit nodes and replace them or just rebuild the lesson UI or the practice UI and restore the student´s work state, whichever works best. I unfortunately don't think the API gives us many more options.

Wolfy7:
With commit https://github.com/GDQuest/learn-gdscript/pull/689/commits/ed582baaed7f05912db90ae3b718daa5908f9d80

Added a CheckButton to the SettingsPopup to enable/disable Lower contrast color for `Label `and `RichTextLabel`.
Defined a Placeholder for "lower contrast color" and "Defaullt color"

Wolfy7:
Hi Nathan.

Are these changes are okay for now? Will you change the "Default color"?

NathanLovato:
Sure, thanks for the bump, I'm adding this to my to do list. The changes look good.

PR #692, opened 2022-10-04
update lesson 18

### LINKED ISSUE #681: Lesson 18, the for-loop is NOT doing the same as the while-loop
The while-loop checks the current position each time, so it works for all cells on the board. It only loops the required amount to reach the right edge.
The for-loop always walks board_size.x times to the right and therefore only works for starting positions on the left edge of the board. If the starting cell is at least one column away from the left edge, it over shoots.

It *could* be remedied by changing the loop to this:

```
for number in range(board_size.x - cell.x - 1):
    cell.x += 1
```

But I doubt you could still call that one a little simpler than the while-loop variant.

ron0studios:
resolves [#681](https://github.com/GDQuest/learn-gdscript/issues/681)

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #

Content amendment

no.

## New feature or change ##

Amended lesson 18 to introduce more clarity, to resolve #681 .

The for-loop would technically not work for different coordinates in the example game, 
but since the player starts at the left of the board it works. This is not explicitly mentioned,
and may confuse some learners, as pointed out in #681 

**What is the new behavior?**

Some clarifications were added, though minor. If there are any others related to #681 , please forward it here, @sgoman. 

**Other information**

### DIFF
--- a/course/lesson-18-for-loops/lesson.tres
+++ b/course/lesson-18-for-loops/lesson.tres
@@ -220,7 +220,7 @@ title = ""
 type = 0
 text = "Rather than constantly checking if the robot reached the end of the board, with the [code]for[/code] loop, we take the board's width beforehand, then move the robot a set amount of times.
 
-The function still works the same. You can execute it below."
+The function still works the same, in this scenario, as our character starts on the opposite side of the board. You can execute it below."
 visual_element_path = ""
 reverse_blocks = false
 has_separator = false


NathanLovato:
The required edits here are bigger than that. See the conversation in https://github.com/GDQuest/learn-gdscript/issues/681. This general part of the lesson could use a bit of a rethink

ron0studios:
ok Im going to try and revamp some more parts of the lesson. I'll have something ready by next week 👍🏽

NathanLovato:
Closing as the text wasn't updated. Don't hesitate to make updates if you'd like to continue working on this pull request, and I'll reopen it.

PR #700, opened 2022-10-10
(NEEDS TESTING) Register macOS shortcuts only on browsers other than Firefox. Fix #698.

### LINKED ISSUE #698: Paste is applied twice in code editor in some browsers
**Describe the bug**
When you copy some code in code editor it will paste code twice.

**To Reproduce**
Steps to reproduce the bug:
1. Go to lesson ,,Drawing a bigger rectangle"
2. Copy first two lines of code with CTRL+C
3. Paste those lines with CTRL+V
4. See the issue

**Expected behavior**
Text should be pasted only once, not twice.

**Screenshots**
2022-10-08 13_04_34-Window

 - Operating System: Windows
 - Browser:  Firefox
 
 Maybe this issue happens also in different lessons. I don't tried it now.

henriquelalves:
I identified the error as Firefox detecting the macOS shortcuts AND the normal shortcuts (not registering the shortcuts makes Firefox behave normally again). I expect the shortcuts not to be necessary on macOS if the user is using Firefox, but I don't have a mac, so this needs testing before merging.

### DIFF
--- a/ui/UICore.gd
+++ b/ui/UICore.gd
@@ -54,14 +54,17 @@ func _ready() -> void:
 	_loading_screen.hide()
 	_welcome_screen.appear()
 
-	# /!\ In the browser we must register macOS shortcuts by hand in Godot 3.
-	# Godot otherwise uses Ctrl-based shortcuts only, making the experience of
-	# mac users so-so.
+	# /!\ In browsers other than FireFox we must register macOS shortcuts by hand 
+	# in Godot 3. Godot otherwise uses Ctrl-based shortcuts only, making the
+	# experience of mac users so-so.
 	#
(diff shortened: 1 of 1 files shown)


NathanLovato:
@lumenwrites could you test this one for us? We need you to check that copy paste works with Cmd C Cmd V in script editors in any browser other than Firefox on mac (safari, chrome...).

To test that you can clone the repo, checkout to this PR, and run this

```
./run prepare:local # downloads specific export templates for this project
./run web:debug # export to html5 and start local server
```

NathanLovato:
I'll revert the macos shortcut as they're included in some browser and not in others, it's something we'd need to solve Godot-side.

Issue #706, opened 2022-10-12
Lesson 11 Time Delta: Rotating Using Delta Practice BUG


DonVonDolo:
When I went to multiply delta to the rotate(2) I left it outside the parenthesis but it still said I was doing it right and congratulated me and sent me on my way to the next practice where I did the same mistake and I was wrong. I didn't realize until I showed a friend who went through the course and he said that should not have said I was right. 

**To Reproduce**
Steps to reproduce the bug:
1. Go to https://gdquest.github.io/learn-gdscript/#course/lesson-11-time-delta/lesson.tres
2. Click on Rotating Using Delta Practice
3. put in the code: 
func process(delta):
     rotate(2) * delta
                                  

**Expected behavior**
Although I did not know at the time, it is supposed to tell me I was wrong and put the * delta inside the parenthesis next to the 2

**Information about your device (please complete the following information):**
 - Operating System: Windows 10
 - Browser: Firefox

![Screenshot_20221012-094406~2](https://user-images.githubusercontent.com/115640954/195359626-f2b2fd0d-82e5-4416-b1d9-5b55d3e343e8.png)

phealy3330:
This also happens with the Linux version of the learn to code app
https://user-images.githubusercontent.com/13488079/196044851-50df31bc-8d04-4a70-883d-9f489644750c.mov

I think possibly the bug is in the test_rotating_character_is_time_dependent() function
https://github.com/GDQuest/learn-gdscript/blob/11730a314919541de40997dad431904dde3c8515/course/lesson-11-time-delta/rotating-using-delta/TestsRotatingDelta.gd#L23

It looks like the check logic does this:
`var has_delta: bool = _body.rfind("delta") > 0`
Does the body of the code have the string delta? Which passes

`var is_in_parentheses: bool = _body.rfind(")")`
Does the body of the code have a close-parenthesis? Which also passes

The second check seems to pass for similar reasons:
https://github.com/GDQuest/learn-gdscript/blob/11730a314919541de40997dad431904dde3c8515/course/lesson-11-time-delta/rotating-using-delta/TestsRotatingDelta.gd#L35

In test_rotation_speed_is_2_radians_per_second()
```
		var has_multiplication_sign: bool = _body.rfind("*") > 0
		var has_two: bool = _body.rfind("2") > 0 and not _body.rfind(".2") > 0
```
The code also has a multiplication sign and a 2. 

[...]

PR #722, opened 2022-10-25
Fix/706 lesson 11 test bug

phealy3330:

- [x ] The commit message follows our guidelines. (Has the prefix fix and linked to #706 on each message)

- For bug fixes and features:
    - [x ] You tested the changes.
 The following test cases were run through the regexp, though not exhaustive is fairly comprehensive:

> rotate(2) #yes
> rotate( 2 ) #yes
> rotate(   2    ) #yes
> rotate(2) #yes
> rotate(2 delta ***   ) #yes
> rotate(2 delta    ) #yes
> rotate(2    ) #yes
> rotate(2.0 delta ***   ) #yes
> rotate(2.0) #yes
> rotate(2.00) #yes
> rotate(2.000000) #yes
> rotate(2.0000000) #yes
> rotate(  2.0 ) #yes
> rotate(2.00  ) #yes
> rotate(2.000000) #yes
> rotate(2.0000000) #yes
> rotate( 2) #yes
> rotate( 2 delta ***   ) #yes
> rotate( 2 delta    ) #yes
> rotate( 2    ) #yes
> rotate( 2.0 delta ***   ) #yes
> rotate( 2.0) #yes
> rotate( 2.00) #yes
> rotate( 2.000000) #yes
> rotate(** delta   2.0000000) #yes
> rotate(     2.0 ) #yes
> rotate(    2.00  ) #yes
> rotate(delta  2.000000) #yes
> rotate( delta 2.0000000) #yes

> rotate() 2 #no
> rotate(2.) #no
> rotate() 2 ) #no
> rotate()2 ) #no
> rotate()2) #no
> rotate(2.01) #no
> rotate(2.00001) #no
> rotate(2.001) #no
> rotate(2.0001) #no
> rotate(200) #no
> rotate(.2) #no
> rotate(.20) #no
> rotate(.200)#no
> rotate(2.1)#no
> rotate(2.5)#no
> rotate(20)#no
> rotate(25)#no
> rotate(2 delta *** 05 50  ) #no
> rotate(2 100    ) #no
> rotate( 1..2    ) #no
> rotate( 12    ) #no
> 

Related issue (if applicable): #706 

Adds two Regular expressions to the Lesson 11 Practice 1 tests so that more edge cases are caught

Not to my knowledge

**Other information**
This does actually exclude the technically correct solution: 
`rotate(2*delta*1)` and any combination that satisfies the mathematical identity property `rotate(  1 * 2 * delta)`, `rotate(0+0+1*delta*2)` etc. 

Because the regular expression that matches for 2 or 2.0 will NOT match if any other numbers are detected, other than trailing zeros in the 2.0 case.

 I think this is an acceptable tradeoff, since we're not dealing with symbolic math software and as anybody doing fancy math equations probably knows what they are doing to push the limits and break the check code. Not many "average" users are going to try something like that.

### DIFF
--- a/course/lesson-11-time-delta/rotating-using-delta/TestsRotatingDelta.gd
+++ b/course/lesson-11-time-delta/rotating-using-delta/TestsRotatingDelta.gd
@@ -28,11 +28,11 @@ func test_rotating_character_is_time_dependent() -> String:
 		var is_in_parentheses: bool = false
 
 		#Regular expression to check if delta is inside parenthesis
-		#I know, regexp is ancient evil magic:
+		#I know, regexp is ancient evil magic, as the quote goes:
 		#  Some people, when confronted with a problem, think
 		#  "I know, I'll use regular expressions" 
 		#  Now they have two problems. 
-		# But sometimes it works. 
+		# But sometimes it does actually work
 
 		var regex_delta_in_parentheses = RegEx.new()
 		#This regex matches if:
@@ -49,6 +49,7 @@ func test_rotating_character_is_time_dependent() -> String:
 
 		#thanks to https://regexr.com/ for the awesome regexp playground to help me figure this out after a ton of trial and error and hair pulling. 
 		
+		#with double backlashes to escape the escapes
 		regex_delta_in_parentheses.compile("rotate\\([^\\)]*delta.*\\)") 
 		var result = regex_delta_in_parentheses.search(_body)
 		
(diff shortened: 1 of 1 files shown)


course/lesson-11-time-delta/rotating-using-delta/TestsRotatingDelta.gd
```diff
@@ -22,20 +22,89 @@ func _prepare() -> void:
 
 func test_rotating_character_is_time_dependent() -> String:
 	if not _has_proper_body:
+		#reverse find for any delta at all
 		var has_delta: bool = _body.rfind("delta") > 0
```
NathanLovato, line 26: You can use type inference in cases like these:

```gd
var has_delta := _body.rfind("delta") > 0
```
We recommend that because the bool type is implied by the expression, and it'll be registered by Godot which will be able to document this variable as being a bool

course/lesson-11-time-delta/rotating-using-delta/TestsRotatingDelta.gd
```diff
@@ -22,20 +22,89 @@ func _prepare() -> void:
...
-		var is_in_parentheses: bool = _body.rfind(")")
+		#set parentheses found false by default
+		var is_in_parentheses: bool = false
 
+		#Regular expression to check if delta is inside parenthesis
+		#I know, regexp is ancient evil magic, as the quote goes:
```
NathanLovato, line 31: Regexes are perfectly fine! Next time could you remove the comments? We can read regexes and the variable name already documents what it does

NathanLovato:
Thank you very much for the pull request, these contributions are much appreciated!

I refactored the code a little bit. It makes a lot of sense to use a regular expression here. The expression's design makes it a little more complicated than it needs to be, but it's fine. It works, so no need to change it at this point.

You can find my changes in the last commit, don't hesitate to add comments if you have any questions

NathanLovato:
Last note regarding commit messages, we use `prefix: user-friendly message here` The prefix is typically `fix:`, `content:`, `improvement:`, `feat:`

For example here: `fix: allow writing decimal number in lesson 11 practice`

Issue #799, opened 2023-03-05
Lesson 21 Strings, last example [step] executes all actions


SchrodingerL10N:
**Describe the bug**
In Lesson 21 Strings, clicking the [step] button executes the full string in one click just like [perform_combo()] does instead of executing one action per step clicked

**To Reproduce**
Steps to reproduce the bug:
1. Go to last example in Lesson 21 Strings
2. Click on [step]
3. See error

**Expected behavior**
After clicking [step] it should execute one action corresponding to the string each time it is clicked

**Screenshots**
If applicable, drag and drop screenshots here to help explain your problem.

**Information about your device (please complete the following information):**
 - Operating System: Windows 11
 - Browser: Chrome Version 110.0.5481.178 (Official Build) (64-bit)

**Additional context**
[gdquest-1677975919049.log](https://github.com/GDQuest/learn-gdscript/files/10890480/gdquest-1677975919049.log)

PR #801, opened 2023-03-05
fix: Lesson 21 - remove Step button from combo example

btstewart21:

- [X] The commit message follows our guidelines.
- For bug fixes and features:
    - [X] You tested the changes.

Related issue (if applicable): #799 

**Other information**
I'm assuming the 'Step' button being added to the final example in Lesson 21 was an accident - the code that looks for `yield()` to determine whether or not to add the 'Step' button to the example was catching the `yield()` that was commented out.


NathanLovato:
Thanks!

Issue #813, opened 2023-04-05
Automated Build System to Check and Integrate Language Translations


NathanLovato:
Issue: We need a build system that can automatically check if translations for a given language are (nearly) complete and integrate those translations into the app. This build script should be able to add or remove languages automatically depending on the completion level of a given language.

Workflow:

1. Extract translation strings from the source application, which produces POT (PO template) files that contain the strings to translate with no translations. To do so, run `i18n/extract.py` from the project's root folder.
2. Move files to the repository where translations are stored: https://github.com/GDQuest/learn-gdscript-translations
3. In the translations repository, run the Python script `sync_translations.py` to update all the PO files which contain the actual translations per language.
4. Commit changes to the repository.
5. Translators update translations using PO files.
6. We integrate the relevant translation files into the app.

We need to ensure that lessons are fully translated (maybe over 90% or 95% translated is OK) before integrating translation files into the app. This feature should be included as part of the build script. The script should also warn about languages with outdated or incomplete translations.

To register languages in the app (show in menu), you can:

1. Copy the folder containing PO files to the i18n/ directory.
2. Add the language code to autoload/TranslationManager.gd, in the SUPPORTED_LOCALES array.

Unfortunately, tooling for PO files is not great, so I suggest implementing our mini-parser to check the completion percentage. A minimal parser is available on this link: https://github.com/GDQuest/learn-gdscript/blob/b53ba8e1cc5da074520b0e061c50c02d2a5fe3fc/i18n/match_and_merge_po_translations.py

It can be used as a base to read entries and calculate the translation percentage. In PO files, entries that need translation have the `fuzzy` keyword. I think that's all we need to check but be sure to double-check if other cases exist.

ml4nC3:
Hi NathanLovato,

Thanks for this. As I'm not familiar with build systems, also considering this is my first open-source contribution, may I ask further guidance on following points ? 
- Do you confirm that this script mission ends on step 4 of your workflow ?
- Also, I understand there is a step 4.5 which outupts translations completion indicators to... the user that runs the script ? Or should it commit some report for everyone else ?
- What should be the path of this script in the project ?
- As you're mentionning Build Systems, is the script going to be executed automatically by something like jenkins or Github Actions or will this be run manually by one of the projects owners during build process ?

*If the script runs locally :*
- should it pull ? 
- If not, I guess some parameter has to be set to target the appropriate folders, at least for the learn-gdscript-translations location
[...]

PR #814, opened 2023-04-20
#813

ml4nC3:

- [ ] The commit message follows our guidelines.
- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #813 

Introducing a script to integrate translations from GD-Learn-Translations if their completion is good enough.

I don't think so...

## New feature or change ##

`extract.py` script and `sync-translations.py` script have to be executed manually. Once done it is difficult to know how well a translation is up to date regarding newly generated POT files.

**What is the new behavior?**
The additional script `check_and_integrate_translations.py` execute sequentially extraction and sync, then looks for untranslated and fuzzy entries in PO files in order to compute some completion indicator. 

If a threshold value is given it then copy translations back to GD-Learn if their completion is above.

**Other information**
! beware this pull request includes a revert on commit https://github.com/ml4nC3/learn-gdscript/commit/297bd15ca33fc23c8b334b7fb9c924f84a5d0d87

This commit seems to be responsible for `msgid` mismatch issues in the application when PO files are integrated back.

### DIFF
--- a/i18n/check_and_integrate_translations.py
+++ b/i18n/check_and_integrate_translations.py
@@ -3,10 +3,39 @@ import sys
 import argparse
 import subprocess
 import shutil
+import glob
 from match_and_merge_po_translations import parse_po_file
+from dataclasses import dataclass
 
 
-def parse_command_line_arguments():
+@dataclass
+class TranslationsData:
+    language_code: str
+    directory_path: str
+    total_strings: int = 0
+    missing_translations: int = 0
+    fuzzy_translations: int = 0
+
+    _completion_rate: int = -1
+
+    def get_completion_rate(self) -> int:
+        """Returns the completion rate of the language with an int between 0 and 100, with 100 representing 100% completion."""
+        if self._completion_rate == -1:
+            todo_count = self.missing_translations + self.fuzzy_translations
+            self._completion_rate = 100 - (100 * todo_count // self.total_strings)
+
+        return self._completion_rate
+
+
+@dataclass
+class Args:
+    translations_path: str
+    threshold: int
+    skip_extract: bool
+    skip_sync: bool
+
+
+def parse_command_line_arguments() -> Args:
     """
     Parses command line arguments and returns a dictionary containing options and arguments.
     """
@@ -39,16 +68,17 @@ def parse_command_line_arguments():
         action="store_true",
         help="Skip the synchronization and merge of PO files with the reference POT files.",
     )
-    return vars(parser.parse_args())
+    return Args(**vars(parser.parse_args()))
 
 
 def main():
     args = parse_command_line_arguments()
 
     I18N_DIRECTORY = os.path.dirname(os.path.realpath(__file__))
-    SOURCE_DIRECTORY = os.path.join(I18N_DIRECTORY, "..")
 
-    if not args["skip_extract"]:
+    if args.skip_extract:
+        print("WARN: Skipping strings extraction and POT files generation...")
+    else:
         print(
             "INFO: Extracting translation strings from learn-gdscript source code and generating POT files..."
         )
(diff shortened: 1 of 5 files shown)


NathanLovato:
Thanks for the work. Got a little refactoring to do, mainly in the future I'd recommend using Python's argparse module if you haven't as it's less error-prone than the old getopt.

For example, this code will handle the help message + validate every argument properly and return you a neat dictionary:

```py
def parse_command_line_arguments():
    """
    Parses command line arguments and returns a dictionary containing options and arguments.
    """
    parser = argparse.ArgumentParser(description="This script performs text extraction from the application and move "
                                                 "generated POT files in the translations project in order to compare "
                                                 "with translations source and output a translation completion "
                                                 "indicator for each language.")
    parser.add_argument("translations_path", help="relative or absolute path to GD-Learn-translations directory.")
    parser.add_argument("-t", "--threshold", type=int, help="Minimum completion percentage value for a language to be integrated. "
                                                            "Language integration is skipped if no value is given.")
    parser.add_argument("-E", "--skip-extract", action="store_true", help="Skip the extraction of strings and POT files generation.")
    parser.add_argument("-S", "--skip-sync", action="store_true", help="Skip the synchronization and merge of PO files with the reference POT files.")
    return vars(parser.parse_args())
```

The dictionary returned by the function looks like this:

```py
{'translations_path': 'test', 'threshold': None, 'skip_extract': False, 'skip_sync': False}
```

Anyway I'll refactor, test, and this should be good to merge.

i18n/check_and_integrate_translations.py
```diff
@@ -0,0 +1,154 @@
...
+        # Moving POT files to translations project
+        print("INFO: Moving POT files to translations folder")
+        pot_files = [file for file in os.listdir('i18n') if file.endswith(".pot")]
+        for pot_file in pot_files:
+            move(os.path.join('i18n', pot_file), os.path.join(translations_path, pot_file))
+    else:
+        print("WARN: Skipping strings extraction and POT files generation...")
+
+    # Updating PO files with sync_translations.py
+    os.chdir(translations_path)
```
NathanLovato, line 92: I recommend not changing the directory in those kinds of scripts for two reasons:

1. It's error-prone. For example, this script changes your directory to another working directory, but it doesn't change it back to the source place you were at the end.
2. Doesn't work if you want to integrate into other workflows or CI.

i18n/check_and_integrate_translations.py
```diff
@@ -0,0 +1,180 @@
...
+
+        return self._completion_rate
+
+
+@dataclass
+class Args:
+    translations_path: str
+    threshold: int
+    skip_extract: bool
+    skip_sync: bool
```
NathanLovato, line 35: This is another change I wanted to highlight. It's personal taste, but I wanted to show this in case you've never used dataclasses - they came in python 3.9 or something. They allow you to quickly make data structures basically, mainly to benefit from better code completion.

Again, personal taste, I just like to get help from the code linter when making typos or doing incorrect assignments (you don't get errors for this with dictionaries).

NathanLovato:
Another note regarding additions to .gitignore: in general if you participate to open source projects maintainers will ask you not to push changes that are specific to your development environment.

Instead of using the shared .gitignore file, you can add exclude patterns to `.git/info/exclude`, which is always local to your copy of the repository.

NathanLovato:
Thank you so much for taking the time to work on this! I'll take it for a spin and make a new release by the end of the week :)

i18n/check_and_integrate_translations.py
ml4nC3: Thanks for this advice, I'll keep it in mind next time I'll develop a system script !

i18n/check_and_integrate_translations.py
ml4nC3: From what I understood it is basically a class with automated getters and setters ? 
Well it is indeed a better solution for a complex data structure.

ml4nC3:
> Got a little refactoring to do, mainly in the future I'd recommend using Python's argparse module if you haven't as it's less error-prone than the old getopt.

Also good to know. I have to admit I saw both solutions in documentation and choose getopt for its simplicity, but the result with arparse is indeed much cleaner. 

> Thank you so much for taking the time to work on this! I'll take it for a spin and make a new release by the end of the week :)

You're welcome, I'm glad I helped ! Don't forget there is still this strange behavior regarding the wrapped strings that resulted in unrecognized translations in the end. 

The script will do it's job, but we might not be quite at the end of the road yet. Let me know how it went with your tests. Maybe you can open another issue if you spot something ? You can get my screenshot and comments if needed. 

I've scrolled into GD-Learning code without much success on spotting where strings are actually used. If you need me to continue I might need a few advices on which scenes or GD scripts to look into.

> I think you said you're French too, right? If you happen to go to Game Camp in June, in Lille, ( https://gamecamp.fr/ ) let me know! My teammate and I will be there

Yes ! I live in Nantes. I'd love to go and meet you there ! Unfortunately I'm pretty sure I won't make it, I have already several work appointments on those dates.

But don't worry, I'll stick around ! This first contribution was a really good experience to me. Helps me keeping a few fingers in true development work. You're refactor is also very interesting. So I'll see if there is another issue that I can give a hand on. Feel free to send suggestions if you have anything in mind, now or later.

Issue #833, opened 2023-07-06
Chapter 18 For Loop quiz


clogan77:
Hint 3  says to use (board_size.y - 1) in the range but range already handles the -1 so its not needed.

![image](https://github.com/GDQuest/learn-gdscript/assets/138819338/960b8e9c-98d9-445b-9693-674ee8038b14)

NathanLovato:
If you use range(board_size.y), the character will overshoot the limits of the game grid. Because, if the board has a height of 5 cells, the robot is already on the first cell, and needs to move 4 cells (board height - 1) to reach the last cell.

So, the hint and practice solution are correct.

PR #834, opened 2023-07-09
Fixed Loop quiz incorrect value in range

Ymanawat:
-  [x] Fixed the hint3 in for_move quiz, mentioned in #833 

Screenshot 2023-07-09 115556


NathanLovato:
Thanks for taking the time to contribute!

Actually, the original hint is correct.

If you use range(board_size.y), the character will overshoot the limits of the game grid. Because, if the board has a height of 5 cells, the robot is already on the first cell, and needs to move 4 cells (board height - 1) to reach the last cell.

Issue #844, opened 2023-07-24
Lesson 28: specifying types error in initial assignment values


theuerc:
**Describe the bug**

[] should be Vector(1,1) in the image below, which shows the initial problem setup. It is unsolvable otherwise.

<img width="1320" alt="Screen Shot 2023-07-24 at 3 54 41 PM" src="https://github.com/GDQuest/learn-gdscript/assets/60138157/2770a3ed-7442-4e29-a60d-49cde099f3e6">

**To Reproduce**
Steps to reproduce the bug:
1. Go to ['Lesson-28 Exercise'](https://gdquest.github.io/learn-gdscript/staging/#course/lesson-28-specifying-types/practice-UdOCQiGj.tres)

**Information about your device (please complete the following information):**
 - Operating System: Mac OS
 - Browser: Chrome

NathanLovato:
Closed by #845

PR #845, opened 2023-07-24
content: fix starting_code for lesson 28

theuerc:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #844

Changes [] to Vector2(1, 1) in the starting code for lesson 28.

No

The starting code has [], but the tests check for Vector2(1, 1). And the solution uses Vector2(1, 1) as well.

**What is the new behavior?**
The starting code has Vector2(1, 1)


NathanLovato:
The intention was to have random values that you have to fix, but I see this makes it so values in the start script are shuffled instead. It's fine by me, thanks for your contribution!

Issue #979, opened 2024-08-21
script with str function call fails


Kevar:
**Describe the bug**
When using the str() function in the first exercise, running the script fails

**To Reproduce**
Steps to reproduce the bug:
1. Go to the first exercise
2. Paste the following script :
`func _ready():`
`	var i = 0`
`	print(str(i))`
3. Run the script
4. The output shows an error

**Expected behavior**
The output should show the text "0"

**Screenshots**
![image](https://github.com/user-attachments/assets/33031ad6-84dc-4772-bb71-b5f2b03dedde)

**Generated log:**
[gdquest-1724254390736.log](https://github.com/user-attachments/files/16694680/gdquest-1724254390736.log)

Wolfy7:
The problem lies in the use of the greedy quantifier .*?, which captures as little as possible, but in your case closes up to the first closing bracket.

(script_replacements)
`\\b(?<command>print)\\((?<args>.*?)\\)`

A simpler solution could be to adapt the expression so that it explicitly searches for the appropriate closing bracket.
See PR https://github.com/GDQuest/learn-gdscript/pull/997 which would do this.

PR #997, opened 2024-10-02
improvement: adapt regex to search for the appropriate closing bracket

Wolfy7:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #979 

## New feature or change ##

see https://github.com/GDQuest/learn-gdscript/issues/979

The generated Script will look like:

```
extends Node2D

onready var _animation_tree := find_node("AnimationTree")

func _ready():
	var i = 0
	MessageBus.print_log([str(i], "WelcomeToTheApp", 6, 1))
	yield(get_tree().create_timer(1.0), "timeout")
	Events.emit_signal("practice_run_completed")

func _run():
	_animation_tree.travel("saying_hi")
```

and due to `[str(i]` it will crash.

**What is the new behavior?**

adapt the regex to search for the appropriate closing bracket

image

**Other information**

### DIFF
--- a/autoload/MessageBus.gd
+++ b/autoload/MessageBus.gd
@@ -18,7 +18,7 @@ var script_replacements := RegExpGroup.collection(
 	{
 		"\\b(?<command>prints)\\((?<args>.*?)\\)":
 		"MessageBus.print_log([{args}], \"{file}\", {line}, {char})",
-		"\\b(?<command>print)\\((?<args>.*?)\\)":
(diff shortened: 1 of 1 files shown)


NathanLovato:
Thanks much, just tested it, looking good. I see that the other script replacements need a change, and actually they all follow the same pattern so the regex replace they use doesn't seem too appropriate. I'll open an issue to log that.

Issue #961, opened 2024-06-14
error leson 16


thijmenb1:
**Describe the bug**
i got the error "oh no! The script has an error, but the script verifier did not catch it" on leson 16 and it sad i need to report it

**To Reproduce**
Steps to reproduce the bug:
1. Go to 'leson 16'
2. Click on 'practice one'
3. type in 'func move_to_bottom():
	while cell.x < board_size.x - 1:
		cell.x += 1
	while cell.y < board_size.y -1:
		cell.y += 1'

**Expected behavior**
i expected to pess the level or ge a error that told me how to fix it

**Screenshots**
![image](https://github.com/GDQuest/learn-gdscript/assets/145684235/9a74014d-d0de-41a9-9d32-16918a86c974)

**Information about your device (please complete the following information):**
 - Operating System: [Windows]
 - Browser: [duckduckgo]

**Additional context**

NathanLovato:
Thank you very much for taking the time to tackle issues! Your contributions are much appreciated. I'm sorry it's taking me so long to review and merge everything; after being sick I have a lot to catch up on.

PR #999, opened 2024-10-08
fix: __guard_counter already defined in the scope

Wolfy7:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #961

## New feature or change ##

Currently the app will crash if you have more then one `while` loop in the `script_text` due to:

image

The generate code (`script_text`) looked like this:

```
func move_to_bottom():
	var __guard_counter := 0
	while cell.x < board_size.x - 1:
		__guard_counter += 1
		if __guard_counter > 100:
			break
		cell.x += 1
	var __guard_counter := 0
	while cell.y < board_size.y -1:
		__guard_counter += 1
		if __guard_counter > 100:
			break
		cell.y += 1
```

**What is the new behavior?**
Change will define the variable `__guard_counter` just once and for other while loops it will be set back to 0.

Now the code looks like:

```
func move_to_bottom():
	var __guard_counter := 0
	while cell.x < board_size.x - 1:
		__guard_counter += 1
		if __guard_counter > 100:
			break
		cell.x += 1
	__guard_counter = 0
	while cell.y < board_size.y -1:
		__guard_counter += 1
		if __guard_counter > 100:
			break
		cell.y += 1
```

**Other information**

### DIFF
--- a/ui/UIPractice.gd
+++ b/ui/UIPractice.gd
@@ -306,7 +306,7 @@ func _validate_and_run_student_code() -> void:
 	# Guard against infinite while loops
 	if "while " in script_text:
 		var modified_code := PoolStringArray()
-		var guard_counter_defined = false
+		var guard_counter = 0
 		for line in script_text.split("\n"):
 			if "while " in line and not line.strip_edges(true, false).begins_with("#"):
 				var indent := 0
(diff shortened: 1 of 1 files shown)


Wolfy7:
I have thought about this again, I think that with `nested while loops` this will not work.

Probably it would be better to add an counter/prefix to the `__guard_counter` to have a separate variable for each while loop.

`__guard_counter1`, `__guard_counter2` and so on.

I will provide a commit with this change later during the day.

NathanLovato:
Thank you very much for taking the time to fix this issue! Adding a suffix or a prefix to the variable sounds like the perfect way to go about it.

Wolfy7:
Yes, without the suffix it would look like:

```
func move_to_bottom():
	var __guard_counter := 0
	while cell.x < board_size.x - 1:
		__guard_counter += 1
		if __guard_counter > 100:
			break
		cell.x += 1
		__guard_counter = 0
		while cell.y < board_size.y -1:
			__guard_counter += 1
			if __guard_counter > 100:
				break
			cell.y += 1
``` 

With the suffix (https://github.com/GDQuest/learn-gdscript/pull/999/commits/809408e3d90e55bcc7eb960feb1659ee4b0b4b42) :

```
func move_to_bottom():
	var __guard_counter0 := 0
	while cell.x < board_size.x - 1:
		__guard_counter0 += 1
		if __guard_counter0 > 100:
			break
		cell.x += 1
		var __guard_counter1 := 0
		while cell.y < board_size.y -1:
			__guard_counter1 += 1
			if __guard_counter1 > 100:
				break
			cell.y += 1
```

Issue #931, opened 2024-03-05
Dyslexia accessible font option


S-Davahan:
It would be really helpful if there was a font selection option that included the opendyslexic font for accessibility

Wolfy7:
@NathanLovato which textes should be affected when a different font is selected ?

Could a solution be to change all `font_data` of all `font_resource` to the new/selected font? 

More or less similiar to `func scale_all_font_sizes(size_scale: int, and_save: bool = true) -> void:`

And should there defined list (dropdown) with fonts or should this dynamic list generated by parsing all fonts (*.ttf, *.otf) in `ui\theme\fonts`

PR #1004, opened 2024-10-11
#931 dyslexia accessible font option

Wolfy7:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #931

This PR is a PoC how the font slection could be look like.

image

image

But things probably needs to change:

- Currently the dropdown (available fonts) is hardcoded
  - should probably dynamic
- It overides every `font_data`
  - Maybe just specific font resources should changed


Wolfy7:
@NathanLovato FYI

NathanLovato:
That's a good start, thanks for this! This is a special accessibility feature for dyslexic people, so I think a checkbox would make the most sense: "Dyslexia Font: [ ]"

And I would make the code to apply the font to theme options named more specifically to match.

Wolfy7:
During startup the `font_path` is stored in `_font_defaults` similiar to the `size`.

When "Dyslexia Font" is enabled all `font_path` will be update to the corresponding (Regular, Bold, Italic) "OpenDyslexic" font.

When "Dyslexia Font" is disabled:
image

When "Dyslexia Font" is enabled:
image

NathanLovato:
Really cool, thanks for the help! There are a few things to refine before integrating this:

- Syntax highlighting is lost in code listings. (fig 1)
- For code, the current font is not monospace so code ends up misaligned. There's a mono variant that could be used, available here: https://antijingoist.itch.io/opendyslexic
- The font looks a lot bigger at any given font size. We should make sure that when applied it looks perceptually about the same size as the default font.
- The settings menu can end up overflowing the app, preventing you from accessing settings. On small displays the buttons to confirm changes will likely be outside the view even at normal font size. The menu could have code to ensure it's centered on the screen and perhaps also a scroll container wrapping all the options to limit its maximum size + code shortcuts to ensure it can be closed with esc. and applied with enter. (fig 2)
- Minor: the sample text in the settings menu does not get updated when applying the font.

The first three items are most important here, the fourth one will be mitigated by the third point


Fig 1:

image


Fig 2:

image

Wolfy7:
Thanks for the feedback.

1. I cant reproduce this, if I test it Syntax highlighting  still works:
image
image

2. Okay, will add the "OpenDyslexicMono-Regular.otf" and use it for code

3. Yes, you right. Will implement this

4. Good point. Will be done

5. Right. will be changed.

NathanLovato:
OK, if the syntax highlighting loss is only happening on my end, it might be a cache issue. Actually, now I remember having an issue like this in the past due to cache problems.

Thank you very much for your diligence!

Wolfy7:
Changes:

- Added OpenDyslexicMono-Regular.otf 
image

- When Dyslexia Font is enabled reduce the font sice by 1/4  `font_resource.size -= (font_resource.size * 0.25)`

- Added a ScrollContainer and Shortcuts "esc" for close and "enter" for apply buttons 
image

- "Sample text" always gets update when changed or applying  font

NathanLovato:
Thanks much for the work and for your patience! I missed the last notification. I found a little thing to tweak but I'll fix it in the main branch directly

Issue #920, opened 2024-02-03
Completing a practice with wrong code.


Mirrekatten:
**Describe the bug**
I did a practice in "20. String, 2. using an array of strings to play a combo"
and messed up the code but still completed the task.

**To Reproduce**
Steps to reproduce the bug:
1. Go to '20. String, 2. using an array of strings to play a combo' practice.
2. Write code.
3. func run():
	combo = ["jab", "jab", "uppercut"]
	for action in combo:
		play_animation("")
4. See completion.

**Expected behavior**
Thought it would fail.

**Information about your device (please complete the following information):**
 - Operating System: [Windows 10]
 - Browser: [Chrome]

NathanLovato:
Thanks! It was solved by Wolfy7 in the contribution linked above

PR #1006, opened 2024-10-16
fix: improve test_use_for_loop to check if the iterator is used

Wolfy7:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #920 

Improvement of `test_use_for_loop` (course\lesson-21-strings\string_array\TestsStringArray.gd) to check that the iterator is used.

## New feature or change ##

See #920 

**What is the new behavior?**

The test will fail with message that the iterator was not used:

image


NathanLovato:
Excellent, just tested, works like a charm, and the feedback messages to the user are solid. Thanks much!

Issue #622, opened 2022-08-13
Error in `13. Conditions/Limited Healing`


nonchip:
**Describe the bug**

Rule number 3 fails incorrectly in Chapter 13 "Conditions / Limited Healing".

**To Reproduce**
To reproduce the bug, answer correctly with:

```
func heal(amount):
	health += amount
	if health > 80:
		health = 80.0
```

**Expected behavior**
The health by amount, capped at 80, to be correctly identified by the rules check.

**Screenshots**
![image](https://user-images.githubusercontent.com/601018/184499452-f66f1c9a-6edd-42f7-973f-10d3fc9738b4.png)

**Information about your device (please complete the following information):**
 - Operating System: linux
 - Browser: vivaldi

PR #1008, opened 2024-10-17
#622 error in 13. conditions limited healing

Wolfy7:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #622 

## New feature or change ##

Currently the test "Health Takes Different Values" will fail with missleading message when an floatingpoint value is used for `health`

Also the test "Health Does Not Go Above 80" will wrongly pass when an floatingpoint values between 80.1 and 80.9 is set:

**What is the new behavior?**

The test "Health Takes Different Values" will pass correctly even if "80" or "80.0" is set for `health`
image

Also test "Health Does Not Go Above 80" will fail correctly

image

**Other information**


NathanLovato:
Thank you!

Issue #1103, opened 2025-07-15
Correct Solution Rejected. Lesson 11 Practice 2 - Moving in Circle Using Delta


RobertOrange:
Does not pass when using a correct solution with semi-colons at the end of lines or using variables in function calls instead of doing math in the function parameter.

**To Reproduce**
Steps to reproduce the bug:
1. Go to lesson 11, problem 2 "Moving in Circle Using Delta"
2. Click on "Solution"
3. Click on "Use Solution"
4. Add Semi-Colons to each line.
5. Click "Run"
6. See error

Alternatively:
1. create variables inside of the function for rotation_speed and move_speed
2. set rotation_speed to 2 * delta
3. set move_speed to 100 * delta

var rotation_speed = 2 * delta
var move_speed = 100 * delta

4. call rotate(rotation_speed)
5. call move_local_x(move_speed)
6. Click "Run"
7. See error

Refer to the screenshots for examples of the above.

**Expected behavior**
Should pass checks but results in errors.

**Screenshots**
<img width="1104" height="512" alt="Image" src="https://github.com/user-attachments/assets/456ed714-acf2-45ca-8419-9c9e7b2440fd" />

<img width="1129" height="511" alt="Image" src="https://github.com/user-attachments/assets/e87a027f-b1df-4656-bcb0-4a7b0f2db826" />

<img width="1132" height="567" alt="Image" src="https://github.com/user-attachments/assets/02e939cc-1585-40ab-b638-5b2082f8e06a" />

**Information about your device (please complete the following information):**
 - Operating System: Windows
 - Browser: Chrome

**Additional context**
Other problems allow the use of semi colons, so it's odd they cause an issue here despite allowing the code to run.
The error messages aren't explicit whether it's the movement speed or the rotation speed that is the issue.

Issue #1085, opened 2025-05-28
10. 2. Circular Movement - Practice


BDCPatrick:
**Describe the bug**
I added the correct Code, it rotates clockwise, but Check says wrong

**To Reproduce**
Steps to reproduce the bug:
1. Write correct code (even solution) and Run

**Expected behavior**
Clearing Check successfully

**Screenshots**
![Image](https://github.com/user-attachments/assets/8d07a82b-4c6e-4841-a956-4cc1d1bc85d1)

**Information about your device (please complete the following information):**
 - Operating System: Windows 11
 - Browser: Chrome

PR #1135, opened 2025-10-22
script_checking/PracticeTester: accept optional trailing semicolons i…

StanleyAlbayeros:
Close https://github.com/GDQuest/learn-gdscript/issues/1103  and https://github.com/GDQuest/learn-gdscript/issues/1085

### DIFF
--- a/script_checking/PracticeTester.gd
+++ b/script_checking/PracticeTester.gd
@@ -19,13 +19,16 @@ var _slice: SliceProperties
 var _test_methods := _find_test_method_names()
 var _code_lines := []
 
+
 # We're not using _init() because it doesn't work unless you define it and call the parent's constructor in child classes. It would add boilerplate to every PracticeTester script.
 func setup(scene_root: Node, slice: SliceProperties) -> void:
 	_slice = slice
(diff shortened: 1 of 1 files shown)


script_checking/PracticeTester.gd
```diff
@@ -87,6 +80,9 @@ func matches_code_line(target_lines: Array) -> bool:
 
 	for line in _code_lines:
 		line = line.replace(" ", "").strip_edges()
+		# Allow optional trailing semicolons so students who add ';' at the end of lines still pass.
+		while line.ends_with(";"):
+			line = line.substr(0, line.length() - 1)
```
NathanLovato: Thank you for another much welcome contribution! Just so you know, there's the rstrip method built in that will remove trailing characters on the right of the string. I'll make the change directly and merge

StanleyAlbayeros: oh awesome! Still learning GDscript so these annotations are super cool, ty, I never know what from the other languages I actually know (python, C, Rust) exists here so I go with the safe option haha  :)

PR #1146, opened 2025-10-26
Fix: fixed error messages in lesson 20 part 2

### LINKED ISSUE #867: Lesson 20, practice 2: the Checks are mixed up
The practice contains 2 checks:
1. Rectangles Do Not Overlap
2. Rectangles Have The Correct Size

If I draw rectangles of proper size without calling jump(), I get OK in the 1st check and a fail in the 2nd check, when it should be vice versa. 

**Screenshot**
gdquest_issue_screenshot-2023-10-03_01-14

**Generated log**
[gdquest-1696284887903.log](https://github.com/GDQuest/learn-gdscript/files/12786938/gdquest-1696284887903.log)

### LINKED ISSUE #1114: Lesson 19 : Rectangles do not overlap (but they do)
**Describe the bug**
The "Rectangles do not overlap" is succesful but the rectangles are overlapping on the screen.
The "Rectangles have the correct size" fails but the rectangles have the correct size.

**To Reproduce**
Steps to reproduce the bug:
1. Go to Lesson 19 > Back to the drawing board
2. Define the run function

```
func run():
	for rect in rectangle_sizes:
		draw_rectangle(rect.x, rect.y)
```
3. Press run
4. See that the "Rectangles do not overlap" test is succesful
5. See that the "Rectangles have the correct size" fails

**Expected behavior**
I would expect the "Rectangles do not overlap" test to fail.
I woudl expect the "Rectangles have the correct size" to be succesful.

**Screenshots**

Image

 - Operating System: Windows
 - Browser: Firefox

**Additional context**
Nothing to add

### LINKED ISSUE #1126: Lesson 19 practice 2. Back to the drawing board -- wrong checks
When creating rectangles I created rectangles of the right size but forgot to add the jump() function so they overlapped. However, the checks said I had rectangles that did not overlap but that they were of the wrong size. So the checks are listed in the wrong order I believe.

mstphayassin:

- [x ] The commit message follows our guidelines.
- For bug fixes and features:
    - [ x] You tested the changes.

Related issue (if applicable):

Close #1126
Close #1114 
Close #867 

## New feature or change ##

Screenshot from 2025-10-25 17-32-01

**What is the new behavior?**

Screenshot from 2025-10-25 19-14-13

**Other information**

### DIFF
--- a/game_demos/DrawingTurtle.gd
+++ b/game_demos/DrawingTurtle.gd
@@ -60,8 +60,8 @@ func move_forward(distance: float) -> void:
 	var new_point := previous_point + Vector2.RIGHT.rotated(deg2rad(turn_degrees)) * distance
 	new_point = new_point.snapped(Vector2.ONE)
 	var is_closed := false
+	# Consider the polygon closed if a coordinate is repeated.
 	if new_point in _points:
-		# consider the polygon closed if a coordinate is repeated.
 		is_closed = true
 	_points.append(new_point)
 	_temp_command_stack.append(


NathanLovato:
Great catch, thank you very much!

NathanLovato:
Actually, this caused a regression in other practices, so I'll revert this one.

NathanLovato:
This is a slightly tricky issue because this turtle does not only draw polygons, but we also need to account for its current state in the practice, like where it's facing. So if we just close the polygon when moving forward, the turtle may not have done a full drawing turn and returned to its original state of facing right or up or something like that which is something we want to check.

I really appreciate you taking a stab at this. I should have just checked more carefully before merging.

Generally speaking, the app and the program could have really used a refactor and thorough integration testing and the likes, though I think someday we'll simply replace it with something new.

I've drafted a new attempt at a fix that delays closing the shape after turns: https://github.com/GDQuest/learn-gdscript/pull/1151

It still needs a little work to account for the turtle changing position though, as students sometimes move the turtle without using the jump() function.

Issue #1100, opened 2025-07-12
Lesson 25.2


Nillaqq:
**Describe the bug**
Displaying the same unit in all 3 cells counts as a solution, when the intended solution is supposed to have 2 robots and 1 turtle. Along with this, if you run the code differently the screen doesn't reset, so if you put a robot where a turtle is, the turtle won't disappear and instead the robot will appear over the turtle.

**To Reproduce**
Steps to reproduce the bug:
1. Go to the "Placing Units On The Board" practice under lesson 25.
2. Write this as your run function:
func run():
	for cell in units:
		place_unit(cell, "robot")
3. Click run and you can continue with an incorrect solve.

Second bug:
1. Go to the "Placing Units On The Board" practice under lesson 25.
2. Use the recommended solution and hit run.
3. Choose to stay in the practice page and replace the run function with this:
func run():
	for cell in units:
		place_unit(cell, "robot")
4. Click run and see how the turtle doesn't disappear, this also works if you spawn turtles instead of robots, because the screen doesn't get reset.
**Expected behavior**
First glitch:
The "All Items Are Displayed" check should be marked as false and you shouldn't get the option to continue.
Second glitch:
When you press the run button, any sprites that are already visible should disappear.

**Screenshots**
<img width="1704" height="942" alt="Image" src="https://github.com/user-attachments/assets/e4b43f3f-4e59-4d55-9912-84d3e982c021" />

<img width="1708" height="641" alt="Image" src="https://github.com/user-attachments/assets/19fd7fab-9969-4120-82d6-3dc2eb36f573" />

**Information about your device (please complete the following information):**
 - Windows
 - Chrome

**Additional context**

PR #1147, opened 2025-10-26
Fix: fixed lesson 25 part 2 test logic

mstphayassin:

- [ x] The commit message follows our guidelines.
- For bug fixes and features:
    - [ x] You tested the changes.

Related issue (if applicable): #1100 

bug fix

no

## New feature or change ##

The test for completing Lesson 25 Part 2 has imprecise logic which allows you to succeed even if the inputted solution is clearly incorrect. 

**What is the new behavior?**
The logic has been adjusted to only accept a solution which matches the expected solution.
all_robots_error_handled

**Other information**
Issue #1100 describes two bugs. This fix only addresses the first bug and not the second one, which involves the screen not clearing as expected.


NathanLovato:
Thank you very much!

PR #1151, opened 2025-10-26
Attempt a fix for polygons - incomplete, needs to account for position

NathanLovato:
This is another attempt at addressing #1114 without affecting other practices. It's currently incomplete because the possibility to change the position of the turtle anytime without going through `jump()` needs to be taken into account as well.


mstphayassin:
@NathanLovato I'm working on an update on this right now and it looks promising so far. Going to double and triple check that it doesn't break any other functionality throughout the app before I create a pull request though.

Issue #1169, opened 2025-11-14
lesson 20.1 checks malfunctioning


mstphayassin:
**Describe the bug**
While messing around with lesson 20.1 to address issue #1125, I got this ridiculous looking result:

**To Reproduce**
Steps to reproduce the bug:
1. Go to lesson 20 exercise 1
2. Replace the body of the `run()` function with `pass`
4. Succeed the "has robot name variable" test 

**Expected behavior**
You should fail both tests in this situation

**Screenshots**

<img width="1079" height="593" alt="Image" src="https://github.com/user-attachments/assets/d2cd6f42-771c-49e9-9636-b998482a27d8" />

**Information about your device (please complete the following information):**
 - Operating System: Linux Mint

**Additional context**
This bug is related to #1125. I'm going to try to work on a solution that addresses both.

Issue #1125, opened 2025-09-15
Lesson 20.1


ghost:
<img width="2554" height="1428" alt="Image" src="https://github.com/user-attachments/assets/8ccad186-1e85-4f7f-91f8-9314a8b138a5" />

PR #1170, opened 2025-11-14
correct logic for tests in 20.1 and provide more descriptive errors

mstphayassin:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue (if applicable): #1125 and #1169 

The tests should all run appropriately now, as seen in these screenshots:
correct_test_1
correct_test_2
correct_test_3

The only case I wasn't sure about was the situation described in #1125 

While this may be confusing, I believe that failing the test in #1125 is not a bug and should be considered the intended functionality.

The reason for this behavior is the use of the `var` keyword within the scope of the run() function. When `var` is used, a new variable is initialized which is only defined in the scope it's initialized in. So, for example:

```Python

# This variable is defined in the global scope.
var robot_name = "INVALID NAME"

func run():
    # This variable is defined only in the run() function.
    var robot_name = "huda"
   
    # Testing what this variable's name is in the scope of the run() function
    print(robot_name) 
    # Result: huda
 

# Later, outside of the run() function

# Testing if you got the right answer:
print(robot_name)
# result: INVALID NAME
```

In the run() function, there is a local variable called robot_name that you initialized that only exists in run(). Meanwhile, there is a global variable, also called robot_name, that is invalid, and it stays invalid because it was never accessed or changed.

All of this to say that I don't see any other way around this other than explicitly directing users *not* to use `var` in their code here. It might be confusing to them and they might wonder why, but it's even more confusing to leave it as is. 

Here's the new functionality in these cases:
weird_tst


NathanLovato:
Thanks! I went a bit beyond explaining just this export code region thingy and wrote some initial docs for the practices in general: https://github.com/GDQuest/learn-gdscript/blob/main/docs/interactive_practices.md

I hope this helps. In short #EXPORT code regions delineate the practice solution. We used to have a bunch of stuff around that (stub support for a sandboxed GDScript editor, multiple files, etc.), but recently I simplified this part of the codebase to just parse those code regions when loading a practice and registering that code as the sample solution for the practice.

Issue #1188, opened 2025-12-22
Replace the lesson resources with a data format easier to edit


NathanLovato:
Currently, all the lesson/course content is stored as Godot resources. We used that at the time to rely on the built-in serialization and loading in the engine, but it gets in the way when it comes to updating content.

This task is about replacing these resources with a more editable format to make maintenance and adding content much easier moving forward.

We use the BBCode formatting of Godot already for the front end and it's a very straightforward format to parse. So I think we'll use BBCode for everything and as a migration path, we'll parse the BBCode to the resources currently used by the application at runtime.

This will also simplify the translation extraction and make it easy to address issues like #835

PR #1191, opened 2025-12-23
Implement replacement bbcode parser for authoring lesson content

NathanLovato:
This is a work in progress to implement #1188 

Making some edits to the existing lessons again it was pretty slow. The point of this is to have an upgrade path to a new format that will replace the Godot resources, get rid of the editor plugin for authoring, and improve iteration speed when working on the content.

We can still have editor tools later on but now it would become a simpler list of lessons and code editor with syntax highlighting for BBCode tags instead of a more complex interface with blocks we drag and drop.


NathanLovato:
Superseded by #1241

PR #1222, opened 2026-02-08
Use debug templates for exports & prevent crashes in verification function

Razoric480:

- For bug fixes and features:
    - [x] You tested the changes.

- Compile the templates as release_debug (https://github.com/Razoric480/custom-godot-builder has already been edited and re-published). This prevents user scripts that cause a runtime error from crashing the whole app.
  - However, just allowing runtime errors will result in the verifier hanging on the 'Verifying student code...' step without the second change.
- Defer adding the student script and calling its _run function so it happens outside of the verification function. This mimics the behaviour that happened in the previous release with the PracticeTester. This changes allows the `try_validate_and_replace_script` to finish without crashing even if there are (will be) runtime errors.

No

## New feature or change ##

Do any lesson and replace the code with something like:

```gdscript
func _ready():
	var array = []
	array.append()
```

The app will crash when running.

**What is the new behavior?**

Do any lesson and replace the code with something like:

```gdscript
func _ready():
	var array = []
	array.append()
```

The app will run the code, even through the runtime errors, without crashing or hanging on verification.

### DIFF
--- a/export_presets.cfg
+++ b/export_presets.cfg
@@ -13,8 +13,8 @@ script_encryption_key=""
 
 [preset.0.options]
 
-custom_template/debug="templates/godot.x11.opt.64"
-custom_template/release="templates/godot.x11.opt.64"
+custom_template/debug="templates/godot.x11.opt.debug.64"
+custom_template/release="templates/godot.x11.opt.debug.64"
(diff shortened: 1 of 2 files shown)


NathanLovato:
Thank you very much!

PR #1232, opened 2026-02-19
Add danish to the list of translations.

jaynom:


NathanLovato:
Thank you for taking the time to make a PR. However, just so you know, you don't add those manually. They're added automatically and programmatically on build if a translation's completion is above a certain threshold - If I recall correctly, something like 90% complete or above.

I will merge because it's harmless and so you're credited. I assume you're the person who's been contributing the Danish translations a couple of days ago. And if so, first, thank you very much for that!

I'm taking note to make a new release when possible so the translations get updated.

PR #1314, opened 2026-05-09
Create a GDScript AST analysis expression API

### LINKED ISSUE #1281: Expose all the GDScript AST and access to as much runtime info as possible
This task is about exposing APIs to get as much information as we can about the student's GDScript code. We've already assessed that we could expose the GDScript AST and probe it from GDScript.

We at least want to expose the AST part for reading at this point. Ideally, we would also like to be able to read runtime debug info of the program, like the actual calculated values and expressions of local variables, if that is possible. that would allow us to test practices from students in a more thorough and seamless way for beginners in the future.

---

Original issue description:

We have this problem where we need to check in the code that the student has written that certain identifiers are there and we rely on regular expressions for this, or sometimes on a partial "mini tokenizer".

In the past we have considered exposing the GDScript parser AST, but decided it was too much work at the time given how much we had to do on this FOSS app. Now that we have had more time, I'd like to revisit this idea and see if it'd be possible for us to expose some APIs to:

- Get the parsed AST
- Read runtime values

There's a whole class of issues that people otherwise encounter that create a bit of friction that can only be solved through parsing and reading values. E.g. if a student writes:

```gdscript
func _process(delta):
    var speed = 500
    move(speed * delta)
```

ideally in a practice test that verifies that a character moves at a certain speed and that the code is time dependent, I would need to verify that delta is used in the correct place and perhaps that the speed is a certain amount. Once we get farther in our curriculum, we just tell students to use properties, but in Learn GDScript, people are too early in their learning for that.

Razoric480:
As outlined in #1281, it is difficult and error prone to use RegEx to analyze code patterns, especially as they get increasingly complex, or if you need to check an entire function or for loop for correctness.

I'm opening this PR to seek out a bit of feedback, since it's an entirely new API to be introduced and consumed.

Fixes #1281 

This PR has two parts:

## C++
https://github.com/Razoric480/godot/tree/raz/expose-parser would get merged into the custom godot engine's main 4.6.x branch we're compiling from and a new release made (can maybe upgrade to 4.6.3 while we're at it, but that's separate.)

This takes all of the parser nodes that are produced by the GDScriptParser class and exposes them as custom GD<NodeType> RefCounted, which can then redirect all of their various outputs via `get_<data>()` function. like `GDIdentifier::get_name()` to get a StringName. This root class can then be grabbed from the GDScriptErrorChecker, calling `get_root_parse_node()` after calling `set_source()`.

While it's possible to directly access them from the `get_root_parse_node()` and analyze the contents, it's not exactly super friendly to use. Ergo:

## GDScript component
This adds the GDScriptASTAnalyzer and the GDExpr API.

The analyzer gives some high level node fetching, like `get_function_named()`, whereas the GDExpr API allows for complex and complete analysis to make sure the code it received matches some expected output by writing a set of nested expressions.

Example simple:
```gdscript
var run_function := _analyzer.get_function_named("run")
	
if not GDExpr.suite(
	GDExpr.for_loop(
		GDExpr.any_identifier(),
		GDExpr.identifier("inventory"),
	)
).matches(run_function):
	return tr("Your code has no for loop. You need to use a for loop to complete this practice, even if there are other solutions!")
```

Example more complex:
```gdscript
var process := _analyzer.get_function_named(&"_process")
var captures := {}
# func _process(delta):
if not GDExpr.function(
	&"_process",
	GDExpr.suite(
		# rotate(2.0 * delta)
		GDExpr.function_call(
			&"rotate",
			GDExpr.multiply(
				GDExpr.literal(2.0),
				GDExpr.captured_identifier(&"delta", captures)
			)
		)
	),
	GDExpr.parameter(
		# capture delta to be used with captured_identifier()
		GDExpr.any_identifier().capture(&"delta", captures)
	)
).matches(process):
	return tr("The rotation speed is not right. The robot should rotate 2 radians per second. Make sure the call looks like this: rotate(2 * %s)." % [captures.delta])
```

Not all nodes are exposed yet but as they are needed, more can be added since it's just a simple abstract inheritance (GDExpr).

### DIFF
--- a/course/lesson-11-time-delta/rotating-and-moving-delta/TestsRotatingMovingDelta.gd
+++ b/course/lesson-11-time-delta/rotating-and-moving-delta/TestsRotatingMovingDelta.gd
@@ -51,39 +51,48 @@ func test_movement_is_time_dependent() -> String:
 
 
 func test_movement_speed_is_correct() -> String:
-	var used_vars := false
-	
 	var process := _analyzer.get_function_named("_process")
 	if not process:
-		return tr("The '_process(delta)' function is missing; did you remove it?")
+		return tr("The _process function is missing; did you remove it?")
 	
-	var process_delta_name := _analyzer.get_function_parameter_name(process, 0)
-	
-	var rotate_call := _analyzer.get_statement_call_named(process, "rotate")
-	if not rotate_call:
-		return tr("Did you use rotate() to make the sprite rotate?")
-	
-	if not GDExpr.function_call("rotate",
-			GDExpr.multiply(GDExpr.literal(2.0), GDExpr.identifier(process_delta_name))
-	).matches(rotate_call):
-		if GDExpr.function_call("rotate", GDExpr.identifier(".*?", true)).matches(rotate_call):
-			used_vars = true
-		else:
-			return tr("The rotation speed is not right. The robot should rotate 2 radians per second. Make sure the call looks like this: rotate(2 * delta).")
-	
-	var move_call := _analyzer.get_statement_call_named(process, "move_local_x")
-	if not move_call:
-		return tr("Did you use move_local_x() to make the sprite move locally?")
-	
-	if not GDExpr.function_call("move_local_x", GDExpr.multiply(GDExpr.literal(100.0), GDExpr.identifier(process_delta_name))).matches(move_call):
-		if GDExpr.function_call("move_local_x", GDExpr.identifier(".*?", true)).matches(rotate_call):
-			used_vars = true
-		else:
-			return tr("The movement speed is not right. The robot should move 100 pixels per second. Make sure the call looks like this: move_local_x(100 * delta).")
-	
-	if used_vars:
+	if GDExpr.suite(
+		GDExpr.any_of(
+			GDExpr.function_call(
+				"rotate",
+				GDExpr.any_identifier()
+			),
+			GDExpr.function_call(
+				"move_local_x",
+				GDExpr.any_identifier()
+			)
+		)
+	).matches(process):
 		return tr("It looks like you stored values in variables before passing them to the functions.") + " " + \
 				tr("That works in GDScript, but we cannot check the values automatically here.") + " " + \
 				tr("Please write the values directly inside the function calls, like this: rotate(2 * delta) and move_local_x(100 * delta).")
+	
+	var process_delta_name := _analyzer.get_function_parameter_name(process, 0)
+	
+	if not GDExpr.suite(
+		GDExpr.function_call(
+			"rotate",
+			GDExpr.multiply(
+				GDExpr.literal(2.0),
+				GDExpr.identifier(process_delta_name)
+			)
+		)
+	).matches(process):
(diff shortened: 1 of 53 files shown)


NathanLovato:
First, thank you for your work! This is looking like it's working well already. On the API side, I will need to test this on concrete exercises to tell you more.

In general, on the usage side, the ability to express an expression as a tree representing an AST, this looks good to me.

One question is, can you run some of your own code through the parser to get the AST and compare against that? I suppose not, that you'd need another file to turn whatever the gdscript parser gives you into your GDExpr objects?

Razoric480:
Nothing prevents you from creating a new error checker and passing in some source code to create an AST tree.
```gdscript
	var temp_checker := GDScriptErrorChecker.new()
	temp_checker.set_source("""extends Node2D

@export var camera_2d: Camera2D

func _ready() -> void:
	set_process(false)
	camera_2d.set_as_top_level(true)


# EXPORT move_and_rotate
func _process(delta):
	rotate(2 * delta)
	move_local_x(100 * delta)
# /EXPORT move_and_rotate


func _run() -> void:
	reset()
	_process(0.0)
	set_process(true)


func reset() -> void:
	rotation = 0.0
	position = Vector2(0, 0)
	set_process(false)
""")
	var root := temp_checker.get_root_parse_node()
	var temp_analyzer := GDScriptASTAnalyzer.new(root)
```

The only thing lacking would be a way to compare two (sub-)trees together. But such a thing could be written.

But yes it would have to be a theoretical complete tree - you wouldn't be able to just AST a single set of statements sans any context around them. e.g. you'd need to, at the very least, have a `func` entrypoint. So the minimum might be:
```gdscript
	var temp_checker := GDScriptErrorChecker.new()
	temp_checker.set_source("""func _process(delta):
	rotate(2 * delta)
	move_local_x(100 * delta)
""")
	var root := temp_checker.get_root_parse_node()
	var temp_analyzer := GDScriptASTAnalyzer.new(root)
```
to be valid. but it'd be valid, since it turns it into a RefCounted

NathanLovato:
It makes sense. GDScript does not allow parsing a standalone expression like `rotate(2 * delta)`? It's one thing that's really nice about the TreeSitter parser. You can parse and analyze any arbitrary expression l;ike the one above, even if technically it would error out in a GDScript file.

which makes me think that that could have been an option for us to expose as tree-sitter produces the parser as a plain C file. But we might as well piggyback on the official GDScript parser to make sure we match it 1:1.

Razoric480:
The closest we have to that is the Expression type, but it doesn't use the GDScript parser. It uses its own math-function-centric mini-language that happens to be able to call functions.

The GDScriptParser expects the root of the tree to be a ClassNode, so it can only accept full fat scripts. They don't have to be files or in the global space (the compiler does but that its own thing) but they have to be classes.

NathanLovato:
Okay! that's perfectly fine. I think we can work with what you have implemented so far.

Xananax:
For expressions, we could just wrap them in a function, then extract the children tokens of the function I suppose.

Issue #847, opened 2023-08-11
Mobile learning


trvrj:
**Is your feature request related to a problem? Please describe.**

I think it's probably not, as the app tells me that my mobile browser screen is too narrow and to use at my own risk. 

**Describe the solution you'd like**

I would enjoy using this during downtime at work on my phone. The issue I'm running into is I'm unable to tap the IDE portion and enter text to add or edit the code. I have tried with Opera Full, Opera Gaming Browser, Chrome, and Samsung's browser with the same issue. It would be great to have mobility functionality so I can learn no matter where I am. 

**Describe alternatives you've considered**

I have used the app on my desktop and it works just fine. 

**Additional context**

NathanLovato:
The app doesn't support mobile platforms, and adding mobile support for this is not planned. We're working on a new course instead that should support mobile platforms a bit better.

trvrj:
I see. Thank you.

Archeia:
Unsure if this is the place to report, I apologize in advance. I just tried this on mobile (Samsung Android) and while the virtual keyboard works, it is impossible to use special characters like parenthesis. When I press those keys it returns nothing.

NathanLovato:
No worries and thanks for the info! In general, the mobile keyboard remains unsupported in the browser. The main way to add mobile support will remain to port and release the application on mobile devices natively. Then in the browser, we could hack it. But it's a fair amount of work + it's possibly going to be a bit brittle/give a so-so experience.

Either way, it's a lot of work and it's not something that we will be able to do soon.

NathanLovato:
Collecting notes on this by @Xananax from #1336. In short, while we're aware there's a simple way to forward on screen keyboard input from javascript to Godot, there are UX problems to solve and a simple code patch is not enough to address this:

- The tab key (for indentation) still gets swallowed but users need a way to indent their code, and it's done with tabs by default in GDScript.
- There's also the size of the on-screen keyboard to account for. It could cover code that users would not be able to reach in some cases and require detaching and moving the keyboard around.

> The reason we didn't implement the keyboard this way is that it lacks support for tab.
> 
> We didn't decide on a solution for this. We considered:
> 
> 1. Automatically convert double-space to tab, but it wouldn't be an obvious feature without some kind of warning and animation. It could also have false positives.
> 2. Automatically convert space at the beginning of a line to tabs. So matching against ^\s, and ^\t+\s. This too though is non-obvious and could defy expectations.
[...]

PR #1336, opened 2026-06-27
Add support for forwarding touch keyboard input to the application

valiantgamingpoint-prog:

- For bug fixes and features:
    - [x] You tested the changes.

Related issue: #847

### DIFF
--- a/html_export/index_template.html
+++ b/html_export/index_template.html
@@ -59,5 +59,31 @@
       const GDQUEST_ENVIRONMENT = {};
     </script>
     <script type="text/javascript" src="bootstrap.js"></script>
+
+    <!-- --- MOBILE KEYBOARD PASS-THROUGH PATCH --- -->
+    <script type="text/javascript">
+      (function() {
+        // We use a passive listener on the window to catch keys as they happen
+        window.addEventListener('keydown', function(event) {
+          // If our GDScript code callback bridge is initialized on the window
+          if (window.godotMobileCallback) {
+            let keyToSend = event.key;
+            
+            if (keyToSend) {
+              // Fire the character across the bridge to codeeditor.gd
+              window.godotMobileCallback(keyToSend);
+              
+              // Prevent standard arrow-key browser scrolling while typing in the box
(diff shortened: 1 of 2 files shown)


Xananax:
Thank you for this PR. It does work well, I tested it on Firefox Android with the FUTO keyboard. I appreciate the attention to details and blocking arrows scrolling.

This said, I am hesitant to pull it into the main branch. The reason we didn't implement the keyboard this way is that it lacks support for tab.

We didn't decide on a solution for this. We considered:

1. Automatically convert double-space to tab, but it wouldn't be an obvious feature without some kind of warning and animation. It could also have false positives.
2. Automatically convert space at the beginning of a line to tabs. So matching against `^\s`, and `^\t+\s`. This too though is non-obvious and could defy expectations.
3. We looked at other applications, and realized almost all coding applications add their own bar on top of the keyboard, with special keys like tab, escape and so on. This is probably the cleanest solution, but requires quite a bit more work. I'm also not sure if can hide/show the bar only when the keyboard is shown/hidden. Finally, while most applications like Termux and co need to present multiple new keys such as escape, tab, tilda, and so on, we really only need tab, so the design might be a little awkward.
4. There might be other solutions, but these are the ones I remember we identified when we discussed this internally.

If this was an application for people who know how to code, then not having any of this would be ok. But since this is an application aimed at people who might have never coded before, leaving them without any obvious way to enter a tab character is problematic.

I'm not convinced by any of the solutions above, to be honest.

I'm open to any suggestion you may have (or anyone else reading).

Note: once we reach an agreement, I'll review your code, but I'd rather not tell you to change anything before we know if we want the code in. I looked at it quickly, it all seems good for me, but if we find a way to make it work, I'll ask you to port the JS code into `bootstrap.js`, following the structure there. I can help you with that when the time comes.

NathanLovato:
I'll be keeping this around for reference but we wouldn't be able to merge this as it is.

The comments look off though that's just a matter of cleaning things up. Most importantly there's what Jad explained and more generally the user experience side of things that's needed to address issue #847. See the last comment in #847. If anyone reads this and would like to participate to this, that would be most welcome.

PR #1346, opened 2026-07-26
Add Nix flake for NixOS/Nix installation

ghost:
This adds a `flake.nix` to the repository root so NixOS and Nix users can install or run the app without downloading the zip manually. No existing files are modified.

- [x] Tested on NixOS 26.05, x86_64-linux, NVIDIA GPU

## What this enables

Run without installing:
```sh
nix run github:GDQuest/learn-gdscript
```

Install persistently:
```sh
nix profile install github:GDQuest/learn-gdscript
```

Or declare it in a NixOS system configuration:
```nix
environment.systemPackages = [
  inputs.learn-gdscript.packages.x86_64-linux.default
];
```

Only x86_64-linux is supported, matching the available release binaries.

## What it does

The flake fetches the Linux release zip, patches the binary to find its system libraries using NixOS's `autoPatchelfHook`, and wraps it so runtime audio and input libraries are available. It also installs a `.desktop` entry and icon so the app appears in application launchers.

## Updating for a new release

Two values to update in `flake.nix` - the version and the hash:

```nix
version = "1.5.3";  # bump to match the new release tag

src = pkgs.fetchurl {
  url = "https://github.com/GDQuest/learn-gdscript/releases/download/${version}/learn-gdscript-${version}-linux.zip";
  hash = "sha256-...";  # update this
};
```

**Getting the new hash:**

With Nix - set hash to an empty string, run `nix build`, and copy the correct hash from the error output:
```nix
hash = "";
```
```
error: hash mismatch ... got: sha256-AvP37Kfiuhhx7K44EGRlo/pu9L+STj6O/3Y+VERCxMI=
```

Without Nix - use curl and openssl:
```sh
curl -sL https://github.com/GDQuest/learn-gdscript/releases/download/1.5.3/learn-gdscript-1.5.3-linux.zip \
  | openssl dgst -sha256 -binary | base64 | sed 's/^/sha256-/'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,22 @@ Also, you can download the app for Windows, macOS, and Linux on Itch.io: https:/
 
 The desktop version can offer better performance and crisper text than the web version.
 
+### Installing on NixOS
+
+If you use NixOS, you can run the app directly without installing it:
+
+```sh
+nix run github:GDQuest/learn-gdscript
+```
+
+Or install it persistently:
+
+```sh
+nix profile install github:GDQuest/learn-gdscript
+```
+
+Only x86_64 Linux is supported via the flake, matching the available release binaries.
+
 ## Feedback, requests, and discussions
 
 We value feedback and bug reports. We will also consider feature requests, especially if they fit our vision and we feel they benefit programming beginners.
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1784796856,
+        "narHash": "sha256-wWFrV5/Qbm+lyt5x20E/bSbfJiGKMo4RCxZV8cl/WZI=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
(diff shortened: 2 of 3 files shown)


NathanLovato:
Thanks for taking the time to try this, but I'm not sure I want to support this. One reason being that the application is not distributed through GitHub releases; you'll have perhaps noticed that the version that you have in this contribution is very outdated. The main recommended way to follow the app is in the browser.

For the few Nix users who might really want the desktop version, isn't there a command that they can use to run the executable? Is there no way for them to download a game build off the internet and run it easily?

ghost:
Makes sense, I'll close this. The browser version covers the use case well, and I agree that pinning to an outdated GitHub release isn't the right approach for a package that's primarily distributed that way.

As for running executables on NixOS, it's genuinely not as seamless as other distros. nix-ld helps with generic binaries but requires knowing which libraries the binary needs and adding them to your config, and steam-run is an option if the user has Steam installed. Neither is particularly user-friendly for a learning app targeting beginners, which is what the flake was trying to address, though I missed that this version was outdated.

NathanLovato:
Thanks for the info and for taking time to contribute!

PR #1356, opened 2026-08-02
Render at the real screen resolution in the browser

### LINKED ISSUE #1183: Add support for changing the resolution and going up to 4k or at least 1440p
It's all in the title. Currently, especially in the HTML5 version, we limit the resolution because there's quite a bit of overhead rendering-wise in the browser, but it'd be nice to unlock that, and at least give people the option to render at a higher resolution and get crisper text.

NathanLovato:
Close #1183

We used to limit the browser resolution to 1080p. I don't remember the reason, I suppose due to the performance limitations of Godot 3 and its WebGL1 renderer. Here I'm unlocking it because you have a much better experience on a display with a resolution over 1080p this way.

I don't get a GPU load higher than on desktop here, the overhead remains mainly on the CPU for me when publishing to the web no more than on desktop. or anything like that. And normally there shouldn't be much difference between rendering with OpenGL on desktop or WebGL in the browser these days. Could you review the changes are correct to get this? I.e. keep Godot render 16:9 centered in the browser window with borders if needed. If it's good I'll roll this in beta to get some broad user testing.


Xananax (approved):
Looks good!

Note: it seems like there should be a way to let the canvas take its size and use less JS, but in my (possibly limited) test, this doesn't actually work

[css_canvas_resize.patch](https://github.com/user-attachments/files/30634439/css_canvas_resize.patch)

The reason being, we need the button to be aligned at the top of the visually rendered application in the canvas context. However, if we let the canvas naturally resize, its reported size adopts the parent's (despite being visually different), which means we do not know where the top boundary of the internally rendered app is to align the HTML-controlled maximize button.

There may be workarounds, but they probably have diminishing returns.

It's also possible that a completely different setup for the entire page allows to use less JS, but it is also probably not worth exploring.

